PR #5, opened 2021-09-14
Initial KubeArmor CLI

daemon1024:


cmd/annotation.go
```diff
@@ -0,0 +1,36 @@
+package cmd
```
nyrahul, line 1: please add license hdr

nyrahul (commented):
I believe the PR is still in early stages to be reviewed?

nyrahul:
Can we add version, build, commit details as part of `karmor version`? The output should be similar to `go version`.
Look at: https://github.com/ahmetb/govvv

install/install.go
```diff
@@ -0,0 +1,198 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func K8sInstaller(c *k8s.Client) error {
+	env := autoDetectEnvironment(c)
+	if env == "none" {
+		return errors.New("unsupported environment or cluster not configured correctly")
+	}
+	fmt.Printf("Auto Detected Environment : %s\n", env)
+	fmt.Print("Service Account ...\n")
```
nyrahul, line 23: Can we make these prints jazzy? For e.g., using emojis similar to what minikube, cilium clients do?
Sorry, I know this sounds a petty work but it has impact.
If you think this could be done later, that is fine too.

version/version.go
```diff
@@ -0,0 +1,41 @@
...
+	"runtime"
+
+	"github.com/kubearmor/kubearmor-client/k8s"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+var version string
+
+func PrintVersion(c *k8s.Client) error {
+	fmt.Printf("karmor version %s %s/%s\n", version, runtime.GOOS, runtime.GOARCH)
```
nyrahul, line 18: we need build-date as well as part of this.

nyrahul (changes requested):

install/install.go
daemon1024: Yes, I was totally planning to add those but if it's fine let's do it in a seperate PR?

version/version.go
daemon1024: Sure I will add that but any particular reasoning for that? I have usually seen only versions and runtime info.

daemon1024:
Ref #6

nam-jaehyun (approved):
looks good to me.

PR #12, opened 2021-10-25
[VM] Added new command to download vm installation script from kvmsoperator

seswarrajan:


vm/vm.go
```diff
@@ -0,0 +1,121 @@
...
+}
+
+var (
+	serviceAccountName = "kvmsoperator"
+	pbClient           pb.HandleCliClient
+)
+
+func initGrpcClient(ip string) error {
+	// Connect to gRPC server
+	grpcClientConn, err := grpc.DialContext(context.Background(), net.JoinHostPort(ip, "32770"), grpc.WithInsecure())
```
nyrahul, line 30: This PORT should be made configurable using an env var.

vm/vm.go
```diff
@@ -0,0 +1,121 @@
...
+	file, err := os.Create(filename)
+	if err != nil {
+		return err
+	}
+	_, err = file.WriteString(scriptData)
+	if err != nil {
+		return err
+	}
+
+	fmt.Printf("VM installation script copied to %s\n", filename)
```
nyrahul, line 61: better to use log in place of fmt.

vm/vm.go
```diff
@@ -0,0 +1,121 @@
+package vm
+
+import (
+	"context"
+	"errors"
+	"fmt"
```
nyrahul, line 6: remove fmt from deps

vm/vm.go
```diff
@@ -0,0 +1,121 @@
...
+		return options.IP, err
+	}
+
+	return clusterIP, err
+}
+
+func FileDownload(c *k8s.Client, options VmOptions) error {
+
+	// Check if kvmsoperator is up and running
+	if _, err := c.K8sClientset.CoreV1().ServiceAccounts("kube-system").Get(context.Background(), serviceAccountName, metav1.GetOptions{}); err != nil {
```
nyrahul, line 90: The kube-system namespace should not be hardcoded. It is possible for kubearmor to be installed as part of some other namespace.

nyrahul (changes requested):
Please find my comments inline. Thanks

vm/vm.go
seswarrajan: sure Rahul.

vm/vm.go
seswarrajan: Other go files use fmt. Should i still carry out with fmt or log ?

vm/vm.go
seswarrajan: I will check for other possible solution.

vm/vm.go
nyrahul: Hey @seswarrajan , the comment seems to be marked resolved without the actual fix.

nyrahul (changes requested):

cmd/vm.go
```diff
@@ -30,6 +30,8 @@ var vmCmd = &cobra.Command{
 func init() {
 	rootCmd.AddCommand(vmCmd)
 	vmCmd.Flags().StringVarP(&vmOptions.IP, "ip", "i", "none", "External IP of kvmsoperator")
-	vmCmd.Flags().StringVarP(&vmOptions.VMName, "vmname", "n", "", "Name of configured vm")
+	vmCmd.Flags().StringVarP(&vmOptions.Port, "port", "p", "none", "Port of kvmsoperator")
+	vmCmd.Flags().StringVarP(&vmOptions.Namespace, "ns", "n", "none", "Namespace of kvmsoperator")
```
daemon1024, line 34: We should default to `kube-system` namespace if no flag input is provided.

protobuf/vm.proto
```diff
@@ -0,0 +1,20 @@
+
```
daemon1024, line 1: Can we add targets in Makefile for protobuf generation?

protobuf/vm.proto
```diff
@@ -0,0 +1,20 @@
...
+syntax = "proto3";
+
+package protobuf;
+
+option go_package="github.com/kubearmor-client/protobuf";
```
daemon1024, line 6:
```suggestion
option go_package="github.com/kubearmor/kubearmor-client/protobuf";
```

Also I wonder if we should keep this protobuf inside the `/vm` directory since this is related to vm subcommand only! What do you think?

Then it would be 
```suggestion
option go_package="github.com/kubearmor/kubearmor-client/vm/protobuf";
```

vm/vm.go
```diff
@@ -0,0 +1,147 @@
...
+	response, err := pbClient.HandleCliRequest(context.Background(), &pb.CliRequest{KvmName: options.VMName})
+	if err != nil {
+		return err
+	} else {
+		if response.Status == 0 {
+			err = writeScriptDataToFile(options, response.ScriptData)
+		} else {
+			return errors.New(response.StatusMsg)
+		}
+	}
```
daemon1024, line 144: Can be simplified to 
```suggestion
	if err != nil {
		return err
	}
	if response.Status == 0 {
		return writeScriptDataToFile(options, response.ScriptData)
	}
	return errors.New(response.StatusMsg)
```

vm/vm.go
```diff
@@ -0,0 +1,147 @@
...
+
+	if options.IP != "none" {
+		return options.IP, err
+	}
+
+	return clusterIP, err
+}
+
+func validateInputParameters(options VmOptions) bool {
```
daemon1024, line 88: We should mark the flags as required, instead of manual validation. Ref https://pkg.go.dev/github.com/spf13/cobra#MarkFlagRequired

That said, I don't think each of these should be required! Can we use some defaults in some cases?

daemon1024 (changes requested):

cmd/vm.go
seswarrajan: Taken care of the same. Default will be kube-system.

vm/vm.go
seswarrajan: Handled

protobuf/vm.proto
seswarrajan: Handled

vm/vm.go
seswarrajan: Handled.. Only vmname is mandatory. Others have some default

protobuf/vm.proto
seswarrajan: Handled. Added vm-protobuf as target

vm/vm.go
nyrahul: This golang file uses both `log` and `fmt`. No point in using both. Lets just use `log`. If the same problem exists elsewhere, lets fix that too.

nyrahul (approved):

PR #26, opened 2021-11-19
[VM Support] Modified code to identify the namespace of kvmservice instead of inputting the same

seswarrajan:


vm/vm.go
```diff
@@ -15,15 +15,15 @@ import (
 )
 
 type VmOptions struct {
```
daemon1024, line 17:
```suggestion
type Options struct {
```

Unrelated to the PR changes, but can we handle this Lint Warning? Thank You.

seswarrajan: Sure @daemon1024 I will handle this.

nyrahul (changes requested):
LGTM. Can you pls fix the lint issues introduced by new code change?

daemon1024:
All of them are about lacking comments w.r.t. Exported Constants. In my opinion, Most exported consts are self-descriptive so we should configure the linter to ignore that. What do you think @nyrahul?

nyrahul:
> All of them are about lacking comments w.r.t. Exported Constants. In my opinion, Most exported consts are self-descriptive so we should configure the linter to ignore that. What do you think @nyrahul?

I think linter is already been lenient it i.e., not failing the GH action. It is a good practice to add a comment for exported consts, funcs. Hence at least for new code it would be good to follow the practice.

nyrahul (approved):

PR #28, opened 2021-11-27
karmor install --image option

nyrahul:
* needed to specify kubearmor test image with `karmor install`
* fixed lint issues

### DIFF
--- a/install/resource.go
+++ b/install/resource.go
@@ -265,7 +265,7 @@ var hostPolicyManagerDeployment = &appsv1.Deployment{
 	},
 }
 
-func generateDaemonSet(env string) *appsv1.DaemonSet {
+func generateDaemonSet(env string, o Options) *appsv1.DaemonSet {
 
 	var label = map[string]string{
 		"kubearmor-app": kubearmor,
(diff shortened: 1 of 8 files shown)


install/resource.go
```diff
@@ -388,7 +388,7 @@ func generateDaemonSet(env string) *appsv1.DaemonSet {
 					Containers: []corev1.Container{
 						{
 							Name:  kubearmor,
-							Image: "kubearmor/kubearmor:latest",
+							Image: o.KubearmorImage, // "kubearmor/kubearmor:latest",
```
daemon1024, line 391: Would it be better to just use the tag from Options?
Something like
```suggestion
							Image: "kubearmor:kubearmor:"+o.KubearmorImage, // "kubearmor/kubearmor:latest",
```
What do you think?

Rest LGTM 👍🏽

nyrahul: The use case for me was to use my own image in my own dockerhub repo which is something like `nyrahul/kubearmor:latest` .. WDYT?

daemon1024 (approved):

PR #37, opened 2021-12-30
Added new commands and modified existing vm commands for vm onboarding, policy enforcement for nonk8s control plane

seswarrajan:


vm/policy.go
```diff
@@ -38,6 +42,54 @@ func PolicyHandling(t string, path string, o PolicyOptions) error {
...
+
+	if resp, err := client.HostPolicy(context.Background(), &req); err == nil {
+		if resp.Status == 1 {
+			fmt.Print("Success")
+		} else {
+			return fmt.Errorf("failed to send policy")
+		}
+	}
+
+	return nil
```
daemon1024, line 59:
```suggestion
	resp, err := client.HostPolicy(context.Background(), &req)
	if err != nil || resp.Status !=1 {
		return fmt.Errorf("failed to send policy %v",err)
	}
	fmt.Print("Success")
	return nil
```
We don't consider the case where err is not nil

vm/policy.go
```diff
@@ -20,8 +24,8 @@ type PolicyOptions struct {
 	GRPC string
 }
 
-//PolicyHandling Function recives path to YAML file with the type of event and emits an Host Policy Event to KubeArmor gRPC Server
-func PolicyHandling(t string, path string, o PolicyOptions) error {
+func sendPolicyOverGrpc(o PolicyOptions, policyEventData []byte) error {
```
daemon1024, line 27:
```suggestion
func sendPolicyOverGRPC(o PolicyOptions, policyEventData []byte) error {
```

vm/policy.go
```diff
@@ -38,6 +42,54 @@ func PolicyHandling(t string, path string, o PolicyOptions) error {
...
+			fmt.Print("Success")
+		} else {
+			return fmt.Errorf("failed to send policy")
+		}
+	}
+
+	return nil
+}
+
+func sendPolicyOverHttp(policyEventData []byte) error {
```
daemon1024, line 62:
```suggestion
func sendPolicyOverHTTP(policyEventData []byte) error {
```

vm/policy.go
```diff
@@ -38,6 +42,54 @@ func PolicyHandling(t string, path string, o PolicyOptions) error {
...
+		return err
+	}
+
+	resp, err := client.Do(request)
+	if err != nil {
+		return err
+	}
+	defer resp.Body.Close()
+
+	respBody, err := ioutil.ReadAll(resp.Body)
```
daemon1024, line 81: `ioutil` is deprecated so let's use `io`
```suggestion
	respBody, err := io.ReadAll(resp.Body)
```

vm/policy.go
```diff
@@ -58,16 +110,11 @@ func PolicyHandling(t string, path string, o PolicyOptions) error {
...
-	if resp, err := client.HostPolicy(context.Background(), &req); err == nil {
-		if resp.Status == 1 {
-			fmt.Print("Success")
-		} else {
-			return fmt.Errorf("failed to send policy")
+	if err = sendPolicyOverHttp(policyEventData); err != nil {
+		// HTTP connection is not active, hence trying to send policy over gRPC
+		if err = sendPolicyOverGrpc(o, policyEventData); err != nil {
+			// Failed to send policy over gRPC
+			return err
```
daemon1024, line 117: We should add some flag for systemd/control plane mode maybe then default to this

daemon1024 (changes requested):

vm/policy.go
```diff
@@ -38,6 +42,54 @@ func PolicyHandling(t string, path string, o PolicyOptions) error {
...
+}
+
+func sendPolicyOverHttp(policyEventData []byte) error {
+
+	timeout := time.Duration(5 * time.Second)
+	client := http.Client{
+		Timeout: timeout,
+	}
+
+	request, err := http.NewRequest("POST", "http://127.0.0.1:8080/policy", bytes.NewBuffer(policyEventData))
```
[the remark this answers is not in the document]
seswarrajan: Yes.. The control plane is always local.

vm/policy.go
seswarrajan: Should be fine. In this case, it will try passing the information over http if it is not present then pass over to grpc. So control plane configuration is not required.

vm/policy.go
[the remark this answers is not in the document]
seswarrajan: IP and port are passed as separate flag options. Review comment addressed

vm/policy.go
seswarrajan: Need to discuss on the same

vm/policy.go
seswarrajan: Need to discuss on the same

seswarrajan: @nyrahul  Please provide your comments on the same.

nyrahul: Yes, we need flag based control ... since failing from HTTP to GRPC can take time i.e. timeouts ... This needs to be handled.

cmd/vm.go
```diff
@@ -88,4 +58,10 @@ func init() {
 		_ = fmt.Errorf("kvm option not supplied")
 	}
 
+	// options for vm generic commands related to HTTP Request
+	vmCmd.Flags().StringVar(&HttpIP, "http-ip", "http://127.0.0.1", "IP of non-k8s control plane")
+	vmCmd.Flags().StringVar(&HttpPort, "http-port", "8080", "IP and port of http request")
+
+	// Create http address
+	HttpAddress = HttpIP + ":" + HttpPort
```
daemon1024, line 66: When HttpAddress is instantiated the values of IP and Port are nil since the flags are not parsed. You need to declare this under `Run` method. This is the reason why you are not getting values as needed.

cmd/vmlabel.go
```diff
@@ -0,0 +1,88 @@
...
+}
+
+// vmLabelDeleteCmd represents the vm add label command for label management
+var vmLabelDeleteCmd = &cobra.Command{
+	Use:   "delete",
+	Short: "delete label for vm k8s/nonk8s control plane",
+	Long:  `delete label for vm k8s/nonk8s control plane`,
+	RunE: func(cmd *cobra.Command, args []string) error {
+		// Create http address
+		httpAddress := "http://" + HttpIP + ":" + HttpPort
```
h3llix, line 47:
```suggestion
	 We can use `net.JoinHostPort` instead of direct concatenation.
```
Ref: https://pkg.go.dev/net#JoinHostPort

cmd/vm.go
```diff
@@ -21,62 +24,29 @@ var vmCmd = &cobra.Command{
...
-
 // vmScriptCmd represents the vm command for script download
 var vmScriptCmd = &cobra.Command{
 	Use:   "getscript",
 	Short: "download vm installation script for nonk8s control plane",
 	Long:  `download vm installation script for nonk8s control plane`,
 	RunE: func(cmd *cobra.Command, args []string) error {
-		if err := vm.GetScript(client, scriptOptions); err != nil {
+		ip := HttpIP
+		isNonK8sEnv := IsNonK8sEnv
```
nyrahul, line 34: There is no point in having a local variable `isNonK8sEnv`? The same change in above places.

nyrahul (changes requested):
Hey @seswarrajan, Can you please handle the lint issue and one comment I provided? Otherwise, it can be be merged.

cmd/vmlabel.go
seswarrajan: @h3llix I already tried with net.JoinHostPort, but not getting an expected output. Will try and share the results.

cmd/vm.go
seswarrajan: Addressed @nyrahul 
That was kept for vaildation by mistake.

cmd/vm.go
```diff
@@ -88,4 +57,11 @@ func init() {
 		_ = fmt.Errorf("kvm option not supplied")
 	}
 
+	// options for vm generic commands related to HTTP Request
+	vmCmd.PersistentFlags().StringVar(&HttpIP, "http-ip", "127.0.0.1", "IP of non-k8s control plane")
+	vmCmd.PersistentFlags().StringVar(&HttpPort, "http-port", "8080", "IP and port of http request")
+	vmCmd.PersistentFlags().BoolVar(&IsNonK8sEnv, "non-k8s", false, "Enable if non-k8s environment/control-plane")
```
nyrahul, line 63: It is better to keep this true by default since we will be using non-k8s mode more frequently. It is painful to type `--non-k8s` with every `karmor` command.
```suggestion
	vmCmd.PersistentFlags().BoolVar(&IsNonK8sEnv, "non-k8s", true, "Enable if non-k8s environment/control-plane")
```

cmd/vm.go
```diff
@@ -88,4 +57,11 @@ func init() {
 		_ = fmt.Errorf("kvm option not supplied")
 	}
 
+	// options for vm generic commands related to HTTP Request
+	vmCmd.PersistentFlags().StringVar(&HttpIP, "http-ip", "127.0.0.1", "IP of non-k8s control plane")
```
nyrahul, line 61:
```suggestion
	vmCmd.PersistentFlags().StringVar(&HttpIP, "http-ip", "127.0.0.1", "IP of kvm-service")
```

cmd/vm.go
```diff
@@ -88,4 +57,11 @@ func init() {
...
+	vmCmd.PersistentFlags().StringVar(&HttpPort, "http-port", "8080", "IP and port of http request")
```
nyrahul, line 62:
```suggestion
	vmCmd.PersistentFlags().StringVar(&HttpPort, "http-port", "8080", "Port of kvm-service")
```

cmd/vmonboarding.go
```diff
@@ -0,0 +1,76 @@
...
+
+	"github.com/kubearmor/kubearmor-client/vm"
+	"github.com/spf13/cobra"
+)
+
+// vmPolicyAddCmd represents the command for vm onboarding
+var vmOnboardAddCmd = &cobra.Command{
+	Use:   "add",
+	Short: "onboard new vm onto nonk8s control plane",
+	Long:  `onboard new vm onto nonk8s control plane`,
```
nyrahul, line 18:
```suggestion
	Short: "onboard new VM onto non-k8s control plane",
	Long:  `onboard new VM onto non-k8s control plane`,
```

Is this only for non-k8s control plane? I thought VM onboarding command remains same regardless of k8s or non-k8s .. right?

cmd/vmonboarding.go
```diff
@@ -0,0 +1,76 @@
...
+
+import (
+	"errors"
+	"net"
+
+	"github.com/kubearmor/kubearmor-client/vm"
+	"github.com/spf13/cobra"
+)
+
+// vmPolicyAddCmd represents the command for vm onboarding
```
nyrahul, line 14:
```suggestion
// vmOnboardAddCmd represents the command for VM onboarding
```

cmd/vmonboarding.go
```diff
@@ -0,0 +1,76 @@
...
+	RunE: func(cmd *cobra.Command, args []string) error {
+		httpAddress := "http://" + net.JoinHostPort(HttpIP, HttpPort)
+		if err := vm.Onboarding("DELETED", args[0], httpAddress); err != nil {
+			return err
+		}
+		return nil
+	},
+}
+
+// vmListCmd represents the command for vm offboarding
```
nyrahul, line 54:
```suggestion
// vmListCmd represents the command for VM listing
```

cmd/vmonboarding.go
```diff
@@ -0,0 +1,76 @@
...
+var vmListCmd = &cobra.Command{
+	Use:   "list",
+	Short: "list configured vms from nonk8s control plane",
+	Long:  `list configured vms from nonk8s control plane`,
```
nyrahul, line 58:
```suggestion
	Short: "list configured VMs",
	Long:  `list configured VMs`,
```
VM list command will remain same regardless of k8s or non-k8s control plane.

nyrahul (changes requested):
Please find my comments inline.
Also please fix the lint warnings.

cmd/vmonboarding.go
seswarrajan: It can be made common @nyrahul 
Currently the implementation is for non-k8s control plane alone. K8s control plane follows kubectl apply commands

cmd/vm.go
seswarrajan: @nyrahul  Reason for default false is that, this utility is used by both k8s and non-k8s for policy enforcement. In the k8s case, for VM direct, the policy enforcement will  happen over GRPC. Hence thought not to disturb the existing flow. Also the use of non-k8s option is required only on policy enforcement and installation script download.

nyrahul (approved):

nyrahul:
@daemon1024 , can you please review this PR? Thanks

vm/onboarding.go
```diff
@@ -0,0 +1,90 @@
...
+
+	respBody, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return "", err
+	}
+
+	return string(respBody), err
+}
+
+func VmList(address string) error {
```
daemon1024, line 44:
```suggestion
func List(address string) error {
```
Exported function is under vm, so `vm.List` is fine. `vm.VmList` seems odd

vm/policy.go
```diff
@@ -58,16 +102,15 @@ func PolicyHandling(t string, path string, o PolicyOptions) error {
...
-			fmt.Print("Success")
-		} else {
-			return fmt.Errorf("failed to send policy")
+	if isNonK8sEnv {
+		// Non-K8s control plane, hence send policy over HTTP
+		if err = sendPolicyOverHTTP(httpAddress, policyEventData); err != nil {
+			return err
+		}
+	} else {
+		// K8sEnv, hence send policy over gRPC
```
daemon1024, line 111: GRPC is  systemd mode which is also non k8s mode, this doesn't seem right

daemon1024, line 111: GRPC is  systemd mode which is also non k8s mode, this doesn't seem right

daemon1024, line 111: gRPC is  systemd mode which is also non k8s mode, this doesn't seem right

daemon1024 (changes requested):

cmd/vm.go
```diff
@@ -87,4 +56,12 @@ func init() {
 	if err != nil {
 		_ = fmt.Errorf("kvm option not supplied")
 	}
+
+	// options for vm generic commands related to HTTP Request
+	vmCmd.PersistentFlags().StringVar(&HttpIP, "http-ip", "127.0.0.1", "IP of kvm-service")
+	vmCmd.PersistentFlags().StringVar(&HttpPort, "http-port", "8000", "Port of kvm-service")
+	vmCmd.PersistentFlags().BoolVar(&IsNonK8sEnv, "non-k8s", false, "Enable if non-k8s environment/control-plane")
```
daemon1024, line 63: Since it is already under the `vm` subcommand. Non k8s flag doesn't really make sense imo. Perhaps using control plane or kvm... Not sure what would be better here but not nonk8s cause nonk8s applies to both systemd and control plane.

vm/policy.go
```diff
@@ -63,17 +106,16 @@ func PolicyHandling(t string, path string, o PolicyOptions) error {
...
-	}
+	if isNonK8sEnv {
+		// Non-K8s control plane with kvmservice, hence send policy over HTTP
+		if err = sendPolicyOverHTTP(httpAddress, policyEventData); err != nil {
+			return err
+		}
+	} else {
+		// Systemd mode, hence send policy over gRPC
+		if err = sendPolicyOverGRPC(o, policyEventData); err != nil {
+			return err
```
daemon1024, line 117: Like I mentioned in flags perhaps naming the flag control plane or something would be better rather than Non k8s.

daemon1024: Also maybe we can have additional checking like if we have a gRPC flag passed down then don't bother checking the flag.

seswarrajan: Since we are using kvm-service, we can name that flag as kvms instead of non-k8s. That will be different. Please suggest

cmd/vm.go
seswarrajan: Since we are using kvm-service, we can name that flag as kvms instead of non-k8s. That will be different. Please suggest

daemon1024: Sure.

daemon1024 (approved):
LGTM 👌

nyrahul (approved):
LGTM

PR #42, opened 2022-02-08
new containerImage field added

nyrahul:
KubeArmor protobufs have changed to added new containerImage field.

karmor protobufs need to be updated to use this field.

Signed-off-by: Rahul Jadhav <nyrahul@gmail.com>


daemon1024 (approved):
Ref https://github.com/kubearmor/KubeArmor/pull/603/
and https://github.com/kubearmor/KubeArmor/pull/596

PR #43, opened 2022-02-10
update kArmor to use stable KubeArmor release instead of latest

Ankurk99:


nyrahul (approved):

daemon1024:
@nyrahul would we want a `--latest` option or it's fine to manually do this with `--image=kubearmor/kubearmor:latest`

daemon1024 (approved):

Issue #40, opened 2022-01-26
filtering options for karmor


nyrahul:
Following filtering options are needed with karmor:
- [ ] `--since=1h`
- [ ] `--namespace=default`
- [ ] `--log=hostlog/containerlog`
- [ ] `--operation=process/file/network`

Syntax should be similar to k8s kubectl syntax where ever applicable.

skant7:
Hi @nyrahul, I was going through the existing code and it seems that some commands like `log` exist as a command with root cmd being `karmor` so the complete command looks like `karmor log --logPath=stdout`, so do we need to get the mentioned filters integrated as flags to the root `karmor` command? 
If so the final cmd looks something like `karmor --since=1h --namespace=default --log=hostlog/container `
Please correct me if I am wrong.

nyrahul:
> Hi @nyrahul, I was going through the existing code and it seems that some commands like `log` exist as a command with root cmd being `karmor` so the complete command looks like `karmor log --logPath=stdout`, so do we need to get the mentioned filters integrated as flags to the root `karmor` command? If so the final cmd looks something like `karmor --since=1h --namespace=default --log=hostlog/container ` Please correct me if I am wrong.

Handling `--namespace` and other filters are much more easy.
Handing `since` is tricky because we do not have historic logs pulled by karmor. Karmor pulls just the instantaneous logs.
Handling `since` might involve some code changes on the base kubearmor code too.

In general, the task at hand is to enable filtering options based on json fields supported and pass this filtering option to the server so that they can be filtered on the kubearmor server itself. (This also means code changes in the base kubearmor code).

(This seems to be a good point to discuss during our next community call.)

Attaching sample telemetry herewith:
```json
{
  "Timestamp": 1643954983,
  "UpdatedTime": "2022-02-04T06:09:43.832838Z",
  "ClusterName": "default",
  "HostName": "ubuntu2004-vagrants",
  "NamespaceName": "explorer",
  "PodName": "mysql-0",
  "ContainerID": "bddbbdb1c00c2092d8b47f3284624cc22e678934d026f300780ee282979e7094",
  "ContainerName": "mysql",
  "HostPID": 934399,
  "PPID": 934391,
  "PID": 177937,
  "UID": 1001,
  "Type": "ContainerLog",
  "Source": "/opt/bitnami/mysql/bin/mysqladmin status -uroot -ppassword",
  "Operation": "File",
[...]

PR #52, opened 2022-03-20
feat: extending filtering options

slayer321:

Add 7 filtering options
```
--namespace
--operation
--logType
--container
--pod
--resource
--source
```
related issue #40 

cc @nyrahul  @daemon1024

### DIFF
--- a/cmd/log.go
+++ b/cmd/log.go
@@ -32,4 +32,11 @@ func init() {
 	logCmd.Flags().StringVar(&logOptions.LogPath, "logPath", "stdout", "Output location for alerts and logs, {path|stdout|none}")
 	logCmd.Flags().StringVar(&logOptions.LogFilter, "logFilter", "policy", "Filter for what kinds of alerts and logs to receive, {policy|system|all}")
 	logCmd.Flags().BoolVar(&logOptions.JSON, "json", false, "Flag to print alerts and logs in the JSON format")
+	logCmd.Flags().StringVar(&logOptions.Namespace, "namespace", "", "Specify the namespace")
+	logCmd.Flags().StringVar(&logOptions.Operation, "operation", "", "Give the type of the operation (Eg:Process/File/Network)")
+	logCmd.Flags().StringVar(&logOptions.LogType, "logType", "", "Log type you want (Eg:ContainerLog/HostLog) ")
+	logCmd.Flags().StringVar(&logOptions.ContainerName, "container", "", "name of the container ")
+	logCmd.Flags().StringVar(&logOptions.PodName, "pod", "", "name of the pod ")
+	logCmd.Flags().StringVar(&logOptions.Resource, "resource", "", "command used by the user")
+	logCmd.Flags().StringVar(&logOptions.Source, "source", "", "binary used by the system ")
 }
--- a/log/log.go
+++ b/log/log.go
@@ -9,17 +9,37 @@ import (
 	"fmt"
 	"os"
 	"os/signal"
+	"regexp"
 	"syscall"
 	"time"
 )
 
+type regexType *regexp.Regexp
+
+var (
+	CNamespace     regexType // CNamespace : It is the regex compiled namespace
+	CLogtype       regexType
+	COperation     regexType
+	CContainerName regexType
+	CPodName       regexType
+	CSource        regexType
+	CResource      regexType
+)
+
 // Options Structure
 type Options struct {
-	GRPC      string
-	MsgPath   string
-	LogPath   string
-	LogFilter string
-	JSON      bool
+	GRPC          string
+	MsgPath       string
+	LogPath       string
+	LogFilter     string
+	JSON          bool
+	Namespace     string
+	LogType       string
+	Operation     string
+	ContainerName string
+	PodName       string
+	Source        string
+	Resource      string
 }
 
 // StopChan Channel
@@ -37,6 +57,40 @@ func GetOSSigChannel() chan os.Signal {
 		os.Interrupt)
 
 	return c
+}
+
+func regexCompile(o Options) error {
+	var err error
+
+	CNamespace, err = regexp.Compile("(?i)" + o.Namespace)
+	if err != nil {
+		return err
+	}
+	CLogtype, err = regexp.Compile("(?i)" + o.LogType)
+	if err != nil {
+		return err
+	}
+	COperation, err = regexp.Compile("(?i)" + o.Operation)
+	if err != nil {
+		return err
+	}
+	CContainerName, err = regexp.Compile("(?i)" + o.ContainerName)
+	if err != nil {
+		return err
+	}
+	CPodName, err = regexp.Compile("(?i)" + o.PodName)
+	if err != nil {
+		return err
+	}
+	CSource, err = regexp.Compile(o.Source)
+	if err != nil {
+		return err
+	}
+	CResource, err = regexp.Compile(o.Resource)
+	if err != nil {
+		return err
+	}
+	return nil
 }
 
 // StartObserver Function
@@ -84,16 +138,21 @@ func StartObserver(o Options) error {
 		fmt.Println("Started to watch messages")
 	}
 
+	err := regexCompile(o)
+	if err != nil {
+		return err
+	}
+
 	if o.LogPath != "none" {
 		if o.LogFilter == "all" || o.LogFilter == "policy" {
 			// watch alerts
-			go logClient.WatchAlerts(o.LogPath, o.JSON)
+			go logClient.WatchAlerts(o)
 			fmt.Println("Started to watch alerts")
 		}
 
 		if o.LogFilter == "all" || o.LogFilter == "system" {
 			// watch logs
-			go logClient.WatchLogs(o.LogPath, o.JSON)
+			go logClient.WatchLogs(o)
 			fmt.Println("Started to watch logs")
 		}
 	}
--- a/log/logClient.go
+++ b/log/logClient.go
@@ -10,6 +10,7 @@ import (
 	"math/rand"
 	"os"
 	"path/filepath"
+	"regexp"
 	"strings"
 	"sync"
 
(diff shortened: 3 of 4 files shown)


log/logClient.go
```diff
@@ -204,6 +205,39 @@ func (fd *Feeder) WatchAlerts(logPath string, jsonFormat bool) error {
 			break
 		}
 
+		if namespace != "" {
```
nyrahul, line 208: Few comments:
* Compile has to be done just once... not on every event.
* avoid code duplication
* add some unit tests

nyrahul (changes requested):
Nice work @slayer321. Few comments:
* Avoid code duplication
* Compile the regex just once
* Add unit tests

log/log.go
```diff
@@ -87,13 +90,13 @@ func StartObserver(o Options) error {
 	if o.LogPath != "none" {
 		if o.LogFilter == "all" || o.LogFilter == "policy" {
 			// watch alerts
-			go logClient.WatchAlerts(o.LogPath, o.JSON)
+			go logClient.WatchAlerts(o.LogPath, o.JSON, o.Namespace, o.LogType, o.Operation)
```
daemon1024, line 93:
```suggestion
			go logClient.WatchAlerts(o)
```
We should prolly now pass the entire option struct instead of manually expanding the entire options struct as function object. What do you think?

log/logClient.go
```diff
@@ -204,6 +205,39 @@ func (fd *Feeder) WatchAlerts(logPath string, jsonFormat bool) error {
...
+
+		if operation != "" {
+			re, err := regexp.Compile(operation)
+			if err != nil {
+				return err
+			}
+			val := re.MatchString(res.Operation)
+			if !val {
+				return nil
+			}
```
daemon1024, line 238: Can we make the regex match case insensitive?

I tried it locally, 
` ./karmor log --operation=Network` works as expected but
` ./karmor log --operation=network` fails.

daemon1024 (changes requested):
Thanks for the PR, Looks like a great start 😄 

Left some comments in addition to what @nyrahul suggested.

log/log.go
slayer321: yeah, we can do that ... looks much more cleaner

log/logClient.go
slayer321: yeah, will implement it

cmd/log.go
```diff
@@ -32,4 +32,11 @@ func init() {
 	logCmd.Flags().StringVar(&logOptions.LogPath, "logPath", "stdout", "Output location for alerts and logs, {path|stdout|none}")
 	logCmd.Flags().StringVar(&logOptions.LogFilter, "logFilter", "policy", "Filter for what kinds of alerts and logs to receive, {policy|system|all}")
 	logCmd.Flags().BoolVar(&logOptions.JSON, "json", false, "Flag to print alerts and logs in the JSON format")
+	logCmd.Flags().StringVar(&logOptions.Namespace, "namespace", "", "Specify the namespace")
+	logCmd.Flags().StringVar(&logOptions.Operation, "operation", "", "Give the type of the operation (Eg:Process/File/Network)")
+	logCmd.Flags().StringVar(&logOptions.LogType, "logType", "", "Log type you want (Eg:ContainerLog/HostLog) ")
+	logCmd.Flags().StringVar(&logOptions.ContainerName, "containerName", "", "name of the container ")
+	logCmd.Flags().StringVar(&logOptions.PodName, "podName", "", "name of the pod ")
+	logCmd.Flags().StringVar(&logOptions.Resource, "resource", "", "command used by the user")
+	logCmd.Flags().StringVar(&logOptions.Source, "source", "", "binary used by the system ")
```
slayer321, line 41: `resource` and `source` usage might not be correct ...
correct me with that

cmd/log.go
```diff
@@ -32,4 +32,11 @@ func init() {
 	logCmd.Flags().StringVar(&logOptions.LogPath, "logPath", "stdout", "Output location for alerts and logs, {path|stdout|none}")
 	logCmd.Flags().StringVar(&logOptions.LogFilter, "logFilter", "policy", "Filter for what kinds of alerts and logs to receive, {policy|system|all}")
 	logCmd.Flags().BoolVar(&logOptions.JSON, "json", false, "Flag to print alerts and logs in the JSON format")
+	logCmd.Flags().StringVar(&logOptions.Namespace, "namespace", "", "Specify the namespace")
+	logCmd.Flags().StringVar(&logOptions.Operation, "operation", "", "Give the type of the operation (Eg:Process/File/Network)")
+	logCmd.Flags().StringVar(&logOptions.LogType, "logType", "", "Log type you want (Eg:ContainerLog/HostLog) ")
+	logCmd.Flags().StringVar(&logOptions.ContainerName, "containerName", "", "name of the container ")
+	logCmd.Flags().StringVar(&logOptions.PodName, "podName", "", "name of the pod ")
```
daemon1024, line 39:
```suggestion
	logCmd.Flags().StringVar(&logOptions.ContainerName, "container", "", "name of the container ")
	logCmd.Flags().StringVar(&logOptions.PodName, "pod", "", "name of the pod ")
```
I believe it's sorta obv that we are passing the respective resource name with the flag. WDYT?

log/logClient.go
```diff
@@ -192,8 +193,21 @@ func (fd *Feeder) WatchMessages(msgPath string, jsonFormat bool) error {
 	return nil
 }
 
+func RegexChecker(filterstring string, res string) (bool, error) {
```
daemon1024, line 196:
```suggestion
func RegexMatcher(filter string, res string, caseSensitive bool) (bool, error) {
```
I think it would be better to add the case insensitivity string to filter as part of the function only

log/logClient.go
```diff
@@ -204,9 +218,62 @@ func (fd *Feeder) WatchAlerts(logPath string, jsonFormat bool) error {
 			break
 		}
 
+		if o.Namespace != "" {
+			val, err := RegexChecker(o.Namespace, res.NamespaceName)
+			if !val {
```
daemon1024, line 223:
```suggestion
			match, err := RegexChecker(o.Namespace, res.NamespaceName)
			if !match {
```
It returns whether it was Matched or not, so I think it would be better to use a variable name that conveys the same.

daemon1024 (changes requested):
If the values/names of particular fields can only be in lower case then it should be fine to make the regex match case insensitive for them since it wouldn't matter what case we provide in filter, the resultant value is only in one case. Does this make sense?

Left some more comments in the review. Thank You.

cmd/log.go
slayer321: yeah

log/logClient.go
slayer321: yeah, we can do that but I think adding the `caseSensitive` parameter to the function will add some code duplication inside the `RegexMatcher`  function  ...
the function will look like
```
func RegexMatcher(filter string, res string, caseSensitive bool)(bool , error){

        if !caseSensitive{
	
	re, err := regexp.Compile(filterstring)
	if err != nil {
		return false, err
	}
	match := re.MatchString(res)
	if !match {
		return false, nil
	}
	return true, nil

	} else{

	re, err := regexp.Compile("(?i)"+ filterstring)
	if err != nil {
		return false, err
	}
	match := re.MatchString(res)
	if !match {
		return false, nil
	}
	return true, nil
	}

}
```
rather than for the current Implementation, we only need to add the two extra strings for the `logType` and `operation` field.WDYT?

daemon1024:
```
if caseSensitive {
 filter = "(?i)"+filter
}
... // Rest of the function remains the same
```

I believe this should be enough, no need to have so much duplication🤔

log/logClient.go
```diff
@@ -192,8 +193,25 @@ func (fd *Feeder) WatchMessages(msgPath string, jsonFormat bool) error {
...
+func RegexMatcher(filter string, res string, caseSensitive bool) (bool, error) {
+
+	if caseSensitive {
+		filter = "(?i)" + filter
+	}
+
+	re, err := regexp.Compile(filter)
+	if err != nil {
+		return false, err
+	}
```
daemon1024, line 205: We are still compiling regex on every alert/log. Ideally we should compile the regex once, so even before the options are passed to WatchAlerts/WatchLogs.

slayer321:
Hey @daemon1024, I have made the required changes. Do check and let me know your feedback.
For case insensitive  of both the fields(`--operation`,`--logType`)  I have directly given the string and removed the boolean part .. as we are compiling the regex before passing it to both the functions

log/log.go
```diff
@@ -39,6 +59,18 @@ func GetOSSigChannel() chan os.Signal {
 	return c
 }
 
+func RegexCompile(o Options) {
+
+	CNamespace, _ = regexp.Compile(o.Namespace)
+	CLogtype, _ = regexp.Compile("(?i)" + o.LogType)
+	COperation, _ = regexp.Compile("(?i)" + o.Operation)
+	CContainerName, _ = regexp.Compile(o.ContainerName)
+	CPodName, _ = regexp.Compile(o.PodName)
```
daemon1024, line 68: Pod, Container and Namespace can be case insensitive too, Refer https://github.com/kubearmor/kubearmor-client/pull/52#pullrequestreview-921879220

log/log.go
```diff
@@ -39,6 +59,18 @@ func GetOSSigChannel() chan os.Signal {
...
+	CSource, _ = regexp.Compile(o.Source)
+	CResource, _ = regexp.Compile(o.Resource)
```
daemon1024, line 70: We should return error handle them, just in case there's a invalid regex provided we should return the error saying the same.

daemon1024 (changes requested):

daemon1024 (approved):

daemon1024 (commented):
Nice Work 👌🏽 Can you handle the linter errors (w.r.t your changes) and squash the commits. Thank You

slayer321:
I'm not able to see any linter error ... It is showing `all checks have passed`

daemon1024:
@slayer321 linter warnings*
You can check them in the [files tab](https://github.com/kubearmor/kubearmor-client/pull/52/files)

nyrahul (changes requested):

nyrahul (approved):
Thanks for the PR. LGTM.

Issue #53, opened 2022-03-23
Observability support for kArmor


Ankurk99:
- [x] support different build options (kubearmor, cilium, discovery)
	- [ ] show build options in the `karmor version`
- [x] Handle karmor observe #54 
	- [x] Support different filters
- [x] Handle `karmor policy discovery` (with filters)
	- #56 
	- #57

wazir-ahmed:
@Ankurk99 @nyrahul Since we're bringing the functionalities we had in the APD's shell scripts to `karmor`, we also need a command which does the job of [dbclear.sh](https://github.com/accuknox/discovery-engine/blob/5df1489d264bbce5455082ac1f972ea807a175d2/scripts/dbclear.sh).

(OR) we could bring a ENV flag (or configuration field) in APD which make the APD clear the DB during startup and start doing the discovery after that.

Ankurk99:
I think adding a command will make more sense as user could themself want to run dbclear even after startup.

nyrahul:
@seswarrajan , i thought we already have a way to handle dbclear with corresponding filtering options.

seswarrajan:
Yes @nyrahul We do support dbclear on insight. 
With **insight --req dbclear** currently supports the same.

wazir-ahmed:
@nyrahul @seswarrajan I checked it. The current `karmor` version doesn't have a command `insight --req dbclear`.

PR #54, opened 2022-03-23
add observability support for karmor

Ankurk99:
ref #53


Makefile
```diff
@@ -27,6 +50,9 @@ clean:
 vm-protobuf:
 	cd $(CURDIR)/vm/protobuf; protoc --proto_path=. --go_opt=paths=source_relative --go_out=plugins=grpc:. vm.proto
 
+insight-protobuf:
```
daemon1024, line 53: I don't think this is needed anymore since we are using protobuf from auto-policy-discovery

Makefile
```diff
@@ -11,9 +11,32 @@ endif
...
+endif
+
+ifneq "$(findstring insight, $(MAKECMDGOALS))" ""
+BUILD_TAG := -tags insight
+endif
+
 .PHONY: build
 build:
-	cd $(CURDIR); go mod tidy; CGO_ENABLED=0 go build -ldflags "-w -s ${GIT_INFO}" -o karmor
+	cd $(CURDIR); go mod tidy; CGO_ENABLED=0 go build -ldflags "-w -s ${GIT_INFO}" -tags "cilium kubearmor insight" -o karmor
```
daemon1024, line 29:
```suggestion
	cd $(CURDIR); go mod tidy; CGO_ENABLED=0 go build -ldflags "-w -s ${GIT_INFO}" -tags "cilium insight" -o karmor
```
I think we won't need a tag to build `kubearmor` related stuff since we should include them by  default.

Also not sure if we support build tags as of now, since the package doesn't seem to be having em

go.mod
```diff
@@ -1,27 +1,152 @@
 module github.com/kubearmor/kubearmor-client
 
-go 1.16
+go 1.17
```
daemon1024, line 3: Let's update to 1.18 maybe

insight/insight.go
```diff
@@ -0,0 +1,74 @@
...
+		ContainerName: o.Containername,
+		ClusterName:   o.Clustername,
+		FromSource:    o.Fromsource,
+		Namespace:     o.Namespace,
+	}
+
+	// create a client
+	conn, err := grpc.Dial(gRPC, grpc.WithInsecure())
+	if err != nil {
+		return nil
```
daemon1024, line 55:
```suggestion
		return err
```

insight/insight.go
```diff
@@ -0,0 +1,74 @@
...
+type Options struct {
+	GRPC          string
+	Labels        string
+	Containername string
+	Clustername   string
+	Fromsource    string
+	Namespace     string
+}
+
+func StartInsight(o Options) error {
```
daemon1024, line 28: Can you add comments for the same to avoid lint errors?

daemon1024 (changes requested):

Makefile
Ankurk99: No, we don't support the build tags yet, I will remove these changes from this PR and raise a separate PR for that.

daemon1024 (approved):

nyrahul (approved):

PR #56, opened 2022-04-01
add policy discovery support

Ankurk99:


discover/discover.go
```diff
@@ -0,0 +1,140 @@
...
+)
+
+// Options Structure
+type Options struct {
+	GRPC   string
+	Format string
+	Policy string
+}
+
+func ConvertPolicy(o Options) error {
```
nyrahul, line 27: Can you fix this lin issue in the new code?

discover/discover.go
```diff
@@ -0,0 +1,140 @@
...
+		return nil
+	}
+	defer conn.Close()
+
+	client := wpb.NewWorkerClient(conn)
+
+	var response *wpb.WorkerResponse
+	response, err = client.Convert(context.Background(), data)
+	if err != nil {
+		return errors.New("could not connect to the server. Possible troubleshooting:\n- Check if discovery engine is running\n- Create a portforward to discovery engine service using\n\t\033[1mkubectl port-forward -n explorer service/knoxautopolicy --address 0.0.0.0 --address :: 9089:9089\033[0m\n- Configure grpc server information using\n\t\033[1mkarmor log --grpc <info>\033[0m")
```
nyrahul, line 56: Thanks for adding this diagnostic message. Very helpful.

discover/discover.go
```diff
@@ -0,0 +1,140 @@
...
+				if o.Format == "json" {
+					arr, _ := json.MarshalIndent(policy, "", "    ")
+					str = fmt.Sprintf("%s\n", string(arr))
+					fmt.Printf("%s", str)
+				} else if o.Format == "yaml" {
+					yamlarr, _ := yaml.Marshal(policy)
+					str = fmt.Sprintf("%s\n", string(yamlarr))
+					fmt.Printf("%s", str)
+				} else {
+					fmt.Printf("Currently supported formats are json and yaml\n")
```
nyrahul, line 85: this should be using log and not fmt.
```suggestion
					log.Printf("Currently supported formats are json and yaml\n")
```

discover/discover.go
```diff
@@ -0,0 +1,140 @@
...
+					break
+				}
+			}
+		}
+	}
+
+	return err
+}
+
+func DiscoverPolicy(o Options) error {
```
nyrahul, line 126: please fix these lint issues.

discover/discover.go
```diff
@@ -0,0 +1,140 @@
...
+	if o.Policy == "" {
+		fmt.Printf("Define policy type by using --policy flag\n")
+	} else if o.Policy == "cilium" {
+		o.Policy = "network"
+	} else if o.Policy == "kubearmor" {
+		o.Policy = "system"
+	} else {
+		fmt.Println("Policy type not recognized.\nCurrently supported policies are cilium and kubearmor\n")
```
nyrahul, line 134: use log.Error()

discover/discover.go
```diff
@@ -0,0 +1,140 @@
...
+			}
+		}
+	}
+
+	return err
+}
+
+func DiscoverPolicy(o Options) error {
+	if o.Policy == "" {
+		fmt.Printf("Define policy type by using --policy flag\n")
```
nyrahul, line 128: Shall we be using default policy type if none is specified?

nyrahul (changes requested):
Thanks for the quick PR. Please find the comments.

discover/discover.go
Ankurk99: Yes we can, but I was confused what policy user will want by default. Should we keep it Kubearmor?

discover/discover.go
Ankurk99: You meant fmt.Errorf?

seswarrajan (changes requested):

discover/discover.go
```diff
@@ -0,0 +1,142 @@
...
+	"errors"
+	"fmt"
+	"os"
+
+	"github.com/rs/zerolog/log"
+
+	wpb "github.com/accuknox/auto-policy-discovery/src/protobuf/v1/worker"
+	"github.com/accuknox/auto-policy-discovery/src/types"
+	"google.golang.org/grpc"
+	"gopkg.in/yaml.v2"
```
daemon1024, line 18:
```suggestion
	"sigs.k8s.io/yaml"
```
We already use this in `vm/policy`, so better to use it here too. It claims better handling of YAML and internally uses `go-yaml` so compatible with it.

nyrahul (approved):

seswarrajan (approved):

PR #57, opened 2022-04-08
add policy discovery support

Ankurk99:


cmd/discover.go
```diff
@@ -0,0 +1,35 @@
...
+		if err := discover.DiscoverPolicy(discoverOptions); err != nil {
+			return err
+		}
+		return nil
+	},
+}
+
+func init() {
+	rootCmd.AddCommand(discoverCmd)
+	discoverCmd.Flags().StringVar(&discoverOptions.GRPC, "gRPC", "", "gRPC server information")
```
seswarrajan, line 28: Why GRPC options alone goes with StringVar and others have StringVarP ?

nyrahul (commented):
I think we are good here. Lets fix the conflicts and merge the PR. Thanks

cmd/discover.go
Ankurk99: the 'P' option allows for shorthand expression, which is not needed in the case of gRPC

Ankurk99:
This PR depends on #56

seswarrajan:
> This PR depends on #56

The PR #56 is merged now. Please resolve conflicts and merge.

seswarrajan (approved):

PR #58, opened 2022-04-11
Added new option fromSource for systempolicy discover

seswarrajan:
Dependent PR : https://github.com/accuknox/auto-policy-discovery/pull/455

### DIFF
--- a/discover/discover.go
+++ b/discover/discover.go
@@ -26,6 +26,7 @@ type Options struct {
 	Namespace   string
 	Clustername string
 	Labels      string
+	Fromsource  string
 }
 
 // ConvertPolicy converts the knoxautopolicies to KubeArmor and Cilium policies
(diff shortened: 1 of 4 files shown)


discover/discover.go
```diff
@@ -26,6 +26,7 @@ type Options struct {
 	Namespace   string
 	Clustername string
 	Labels      string
+	Fromsource  string
```
Ankurk99, line 29: FromSource  string

seswarrajan: The naming convention is **Fromsource**. 
Need to use the same.

daemon1024: I would prefer both of em to be updated to `FromSource` unless there's some other limitation

daemon1024: > But we are maintaining this naming convention for all other parameters as well

Only in the worker protobuf in which these fields were added recently. In general the type system in autopol seems to be camelCase only. But yeah it's fine if we want to keep following this.

daemon1024 (approved):

PR #59, opened 2022-04-12
feat: add limit flag

slayer321:

add limit flag
```
--limit
```

related issue #40 

cc @nyrahul @daemon1024

### DIFF
--- a/log/logClient.go
+++ b/log/logClient.go
@@ -18,6 +18,9 @@ import (
 	"google.golang.org/grpc"
 )
 
+var Limitchan chan bool
+var i uint32
+
 // ============ //
 // == Common == //
 // ============ //
@@ -64,6 +67,9 @@ type Feeder struct {
 	// server
 	server string
 
+	//limit
+	limit uint32
+
 	// connection
 	conn *grpc.ClientConn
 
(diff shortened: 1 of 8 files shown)


daemon1024 (commented):
I am a bit skeptical about this approach. Would Atomic Counters be more feasible here and can avoid this much refactoring.

log/logClient.go
```diff
@@ -202,124 +210,143 @@ func regexMatcher(filter *regexp.Regexp, res string) bool {
...
 			}
+			_ = watchAlertsHelper(res, o)
 
-			str = str + fmt.Sprintf("Result: %s\n", res.Result)
 		}
+		Limitchan <- true
+
+	} else {
+		for fd.Running {
+			res, err := fd.alertStream.Recv()
```
nyrahul, line 343: Is the possible to remove this code duplication ... I mean `Recv() ... watchAlertsHelper()`

nyrahul (approved):
LGTM. One comment provided inline. Please check if it can be handled.

nyrahul:
Can you please fix the conflict by rebasing and please check the review comment ? Thanks

log/logClient.go
slayer321: I think currently only code duplication section is this one 
```
res, err := fd.logStream.Recv()
if err != nil {
	fmt.Printf("Failed to receive an alert (%s)\n", err.Error())
	break
}
_ = WatchLogsHelper(res, o)
```

I tried to add it in another function ... but it didn't help

daemon1024 (approved):

PR #62, opened 2022-04-14
use random grpc port

nthnieljson:
log/log.go : use random grpc port

previously we can't use random grpc port

now we can by reading the port number from pid file which is located in /opt/kubearmor

Signed-off-by: Nathaniel Jason <nthnieljson@gmail.com>

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -11,7 +11,7 @@ replace (
 )
 
 require (
-	github.com/accuknox/auto-policy-discovery/src v0.0.0-20220909105533-27ee0d74dfc3
+	github.com/accuknox/auto-policy-discovery/src v0.0.0-20221004060846-9c120a7390e8
 	github.com/blang/semver v3.5.1+incompatible
 	github.com/cilium/cilium v1.10.13
 	github.com/clarketm/json v1.17.1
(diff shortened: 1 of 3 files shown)


nthnieljson:
this is addressing the issue on https://github.com/kubearmor/KubeArmor/issues/623 to use random grpc port number

daemon1024:
Depends on https://github.com/kubearmor/KubeArmor/pull/672

log/log.go
```diff
@@ -103,7 +103,12 @@ func StartObserver(o Options) error {
 		if val, ok := os.LookupEnv("KUBEARMOR_SERVICE"); ok {
 			gRPC = val
 		} else {
-			gRPC = "localhost:32767"
+			fileContent, err := os.ReadFile("/opt/kubearmor/kubearmor.pid")
+			if err != nil || string(fileContent) == "" {
+				gRPC = "localhost:32767"
+			} else {
+				gRPC = fmt.Sprintf("localhost:%s", string(fileContent))
+			}
```
daemon1024, line 111:
```suggestion
			else if fileContent, err := os.ReadFile("/opt/kubearmor/kubearmor.pid"); err == nil || string(fileContent) != "" {
				gRPC = fmt.Sprintf("localhost:%s", string(fileContent))
			} else {
				gRPC = "localhost:32767"
			}
```
ig this would better represent the sequence of checking and assignment. WDYT?

log/log.go
```diff
@@ -102,8 +102,10 @@ func StartObserver(o Options) error {
 	} else {
 		if val, ok := os.LookupEnv("KUBEARMOR_SERVICE"); ok {
 			gRPC = val
-		} else {
+		} else if fileContent, err := os.ReadFile("/opt/kubearmor/kubearmor.pid"); err != nil || string(fileContent) == "" {
```
nyrahul, line 105: Few things:
1. we should check for the presence of this file only in non-k8s env.
2. What happens if a stray pid file lingers after the kubearmor is uncleanly stopped?

nyrahul (commented):
Thanks @nthnieljson for handling this. Mostly LGTM. Please check if comments inline.

log/log.go
Ankurk99: >     1. we should check for the presence of this file only in non-k8s env.
> 
>     2. What happens if a stray pid file lingers after the kubearmor is uncleanly stopped?
@nthnieljson Can you please take a look at it?

nthnieljson: 1. this is the same behavior as before, if there is no file (in non-k8s env), it will chooses the default value, which is "localhost:32767"
2. Then it will use the previous port value, this would cause problem, I am not really sure how to address this, need your input and help.

daemon1024:
Can you please rebase the code with upstream changes and resolve the conflicts?

Rest LGTM

PR #71, opened 2022-05-11
feat: add selector flag to logs

slayer321:

add selector flag
```
-l | --selector 
```

related issue #40 

cc @nyrahul , @daemon1024

### DIFF
--- a/log/logClient.go
+++ b/log/logClient.go
@@ -6,6 +6,7 @@ package log
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"math/rand"
 	"os"
@@ -211,6 +212,14 @@ func regexMatcher(filter *regexp.Regexp, res string) bool {
 }
 
 func watchAlertsHelper(res *pb.Alert, o Options) error {
+
+	if len(o.Selector) != 0 {
+		val := selectLabels(o, res.Labels)
+		if val != nil {
+			return nil
+		}
+	}
+
 	if o.Namespace != "" {
 		match := regexMatcher(CNamespace, res.NamespaceName)
 		if !match {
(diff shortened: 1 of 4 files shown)


log/logClient.go
```diff
@@ -491,3 +508,39 @@ func (fd *Feeder) DestroyClient() error {
...
+	} else if len(o.Selector) >= 2 {
+
+		for _, val := range o.Selector {
+			if val == res.Labels {
+				return nil
+			}
+		}
+
+	}
+	return errors.New("Not found any flag")
```
nyrahul, line 527:
```suggestion
	for _, val := range o.Selector {
		if val == res.Labels {
			return nil
		}
	}
	return errors.New("Not found any flag")
```

log/logClient.go
```diff
@@ -491,3 +508,39 @@ func (fd *Feeder) DestroyClient() error {
...
+}
+
+func selectLabelsLogs(o Options, res *pb.Log) error {
+
+	if len(o.Selector) <= 1 {
```
nyrahul, line 532: Not sure about why we need to explicitly add a check for this. Check the proposed code for selectLabelsAlerts

log/logClient.go
```diff
@@ -491,3 +508,39 @@ func (fd *Feeder) DestroyClient() error {
...
+			if val == res.Labels {
+				return nil
+			}
+		}
+
+	}
+	return errors.New("Not found any flag")
+}
+
+func selectLabelsLogs(o Options, res *pb.Log) error {
```
nyrahul, line 530: can we have a single function `selectLabels(o Options, res *pb.Labels)` ?

nyrahul (changes requested):

nyrahul (changes requested):
All good. Can you squash the changes into single commit? Thanks

nyrahul (approved):

PR #72, opened 2022-05-19
Adding request option for observe/dbclear with insight

seswarrajan:


Ankurk99:
I tried running with 'dbclear' flag but it's unable to connect to the grpc server (working fine for observe though). Can you please take a look?

PR #74, opened 2022-05-20
add unorchestrated docker observability logs support

Ankurk99:
support containerized workloads and add summarized view


log/logClient.go
```diff
@@ -271,15 +272,24 @@ func watchAlertsHelper(res *pb.Alert, o Options) error {
...
+		if _, err := os.Stat("/var/run/docker.pid"); err == nil {
+			dockerenabled = true // docker is running
+		}
+
+		if !dockerenabled {
+			str = str + fmt.Sprintf("Cluster Name: %s\n", res.ClusterName)
+			str = str + fmt.Sprintf("Host Name: %s\n", res.HostName)
+		}
 
 		if res.NamespaceName != "" {
```
daemon1024, line 285: NamespaceName should be empty for our docker container logs right?

log/logClient.go
```diff
@@ -417,15 +428,24 @@ func WatchLogsHelper(res *pb.Log, o Options) error {
 
 		str = fmt.Sprintf("== Log / %s ==\n", updatedTime)
 
-		str = str + fmt.Sprintf("Cluster Name: %s\n", res.ClusterName)
-		str = str + fmt.Sprintf("Host Name: %s\n", res.HostName)
+		// Check if docker daemon is running
+		if _, err := os.Stat("/var/run/docker.pid"); err == nil {
+			dockerenabled = true // docker is running
+		}
```
nyrahul, line 434: It is best to wrap this check in a function and just use that function. In the future, we will use more container runtimes and the internal check would differ.

Also it would be best to use some other name in place of `dockerenabled` so that we can use it for any container runtime.

nyrahul (changes requested):

log/logClient.go
Ankurk99: No, we have set "container_namespace" for un-orchestrated containers

Ankurk99:
I have removed summary handling from this PR, will handle that separately.

daemon1024 (approved):

nyrahul (approved):
LGTM

PR #82, opened 2022-06-12
Synched with /vmlist response format changes in kvm-service

wazir-ahmed:
* **Merge this PR after https://github.com/kubearmor/kvm-service/pull/59 is merged**
```
```


vm/onboarding.go
```diff
@@ -8,12 +8,23 @@ import (
...
 	"path/filepath"
+	"strconv"
+	"strings"
 	"time"
 
 	tp "github.com/kubearmor/KVMService/src/types"
 	"sigs.k8s.io/yaml"
 )
 
+type KVMSEndpoint struct {
```
seswarrajan, line 19: @wazir-ahmed  Kindly fix this linter warning. Add comment for this func

vm/onboarding.go
```diff
@@ -43,17 +54,30 @@ func postHTTPRequest(eventData []byte, vmAction string, address string) (string,
...
+	var endpoints []KVMSEndpoint
 
 	vmlist, err := postHTTPRequest(nil, "vmlist", address)
 	if err != nil {
 		fmt.Println("Failed to get vm list")
 		return err
 	}
 
-	if vmlist == "" {
+	err = json.Unmarshal([]byte(vmlist), &endpoints)
```
seswarrajan, line 65: @wazir-ahmed Instead of using encoding/json import, kindly use clarketm/json formatting. 
This import ignores all empty elements in DB. 

You can refer to https://github.com/kubearmor/kubearmor-client/blob/c9431d1b17247692627f6ed8b3e8af6a1b2fc2ce/discover/discover.go#L12

seswarrajan (changes requested):

vm/onboarding.go
wazir-ahmed: I'm only doing `json --> struct` conversion here. So the standard `json` library works fine in this case.

vm/onboarding.go
wazir-ahmed: Removed the `struct` definition from here and directly used the struct from `github.com/kubearmor/kvm-service/src/types` package.

seswarrajan (approved):

PR #83, opened 2022-06-13
Support input files that contain multiple VM host/network policies

wazir-ahmed:

### DIFF
--- a/vm/policy.go
+++ b/vm/policy.go
@@ -11,6 +11,7 @@ import (
 	"net/http"
 	"os"
 	"path/filepath"
+	"strings"
 	"time"
 
 	v2 "github.com/cilium/cilium/pkg/k8s/apis/cilium.io/v2"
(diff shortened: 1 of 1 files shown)


vm/policy.go
```diff
@@ -107,59 +110,71 @@ func PolicyHandling(t string, path string, o PolicyOptions, httpAddress string,
...
-		}
+		var hostPolicy tp.K8sKubeArmorHostPolicy
+		var networkPolicy v2.CiliumNetworkPolicy
+		var policyEvent interface{}
 
-	}
+		if k.Kind == KubeArmorHostPolicy {
+			err = json.Unmarshal(js, &hostPolicy)
+			if err != nil {
+				return err
```
nyrahul, line 137: error handling in this function is tricky!
What happens if one of the policy unmarshal fails and other were ok. All the policies before the failed policies were applied.

seswarrajan: Instead we can log the error and continue.

seswarrajan (changes requested):

vm/policy.go
wazir-ahmed: > All the policies before the failed policies were applied.

 @nyrahul Yes. This is similar to how `kubectl` works. If we have 5 policies, out of which the first three policies are good and the fourth one is invalid, then `kubectl` will apply the first three policies successfully and throw error while processing the fourth one.

wazir-ahmed: @seswarrajan The error returned from this function will be printed in the user console.

nyrahul (approved):
LGTM

Issue #19, opened 2021-11-09
kArmor probe utility


daemon1024:
Utility to analyse configuration and supported KubeArmor features in the current environment

List of things to probe into
- [x] Host Security Polices
- [x] Pod Security Policies
- [x] Audit Mode
- [x] Enforcement Mode
	- [x] AppArmor
	- [x] SELinux
	- [x] BPF-LSM
- [x] Need a probe-check to see if the kernel-headers are present in the host ([ref](https://github.com/kubearmor/KubeArmor/issues/647), Issue `module kheaders not found in modules.dep Unable to find kernel headers`.)
- [x] Check Pods being handled about KubeArmor

- [x] Check Which policies are applied to which pods!

sumit-158:
Hi , I want to participate in LFX Mentorship currently I'm kinda beginners to go and k8s can you help me out how to get involved in this project or any suggestion ?

SaarthakMaini:
Hello,
I am currently a beginner in Go and Kubernetes
I would also like to contribute to the project under LFX Mentorship Program
May I also follow the above resources and apply for the project?
Thank you!

anutosh491:
Hello @daemon1024  , hope you're doing well , this is Anutosh here from India . 

I'm an open source enthusiast and I am keen to take part in the LFX Mentorship program for the summer term. The project and the library interests me. But being new to the project , I would be glad if you could syggest any relevant resources/links I should be going through as a beginner for getting to know the project and the library better . Thank you !

AnuragThePathak:
Hi, I'm Anurag Pathak, a full-stack developer and B. Tech. 2nd-year CSE student. I have been working with Kubernetes and Golang for the last 2-3 months and contributing to Brigade and Argo-workflow. While looking for some good projects which interest me, I found KubeArmor and eventually came to this issue. I would like to work on it in LFX Mentorship Summer 2022.

SaarthakMaini:
Hello,
 
I have gone through the required pre-requisites and resources as stated above.
I also saw this video: 

https://www.youtube.com/watch?v=A4HwqCRjyHE&t=221s

to learn about KubeArmor and eBPF

What I understand is that we have to create and change yaml files in the policy-templates repository. Am I correct?

If possible, please point me in the direction from where I could get started on this issue and what else I have to do in this exactly, and I will take it on from there

Thank you very much!

shreemaan-abhishek:
[...]

PR #86, opened 2022-06-15
kArmor probe utility #19

Essietom:
Utility to analyse configuration and supported KubeArmor features in the current environment

Previously, karmor doesn't check if a the host machine on which kubearmor will be installed supports audit, this utility will give users headsup as to weather kubearmor will be supported on their machine or not.

This update allows karmor to  check if audit/observability is supported on a host machine before installation of kubearmor is done

Fixes: #19  (partially)


cmd/probe.go
```diff
@@ -0,0 +1,26 @@
...
+
+package cmd
+
+import (
+	"github.com/kubearmor/kubearmor-client/probe"
+	"github.com/spf13/cobra"
+)
+
+// versionCmd represents the get command
+var probeCmd = &cobra.Command{
```
daemon1024, line 12: Let's update the comment to match the the command

probe/probe.go
```diff
@@ -0,0 +1,63 @@
...
+import (
+	"fmt"
+	"syscall"
+    "strings"
+	"golang.org/x/mod/semver"
+
+)
+
+func PrintProbeResult() error{
+	AutoDetectKernelVersion, err := autoDetectKernelVersion()
```
daemon1024, line 16:
```suggestion
	hostKernelVersion, err := detectHostKernelVersion()
```

These names would make more sense imo. WDYT?

probe/probe.go
```diff
@@ -0,0 +1,63 @@
...
+	if err != nil {
+		return nil
+	}
+	if AutoDetectKernelVersion == false {
+		fmt.Printf("**Host** \n")
```
daemon1024, line 21: What I wanted to represent with `**Host**` is that we print Host in a **bold** or some _different_ way.

daemon1024 (changes requested):
Thanks for the PR, Looks like a great start :smile:

Left some comments inline.
Let's either 
- detect Enforcement/LSM for host
or 
- detect Observability Support in K8s environement

in the same PR. I will leave that decision to you :)

cmd/probe.go
Essietom: yes, It makes sense to have the enforcement check for the host to be in the same PR.

probe/probe.go
```diff
@@ -0,0 +1,87 @@
...
+        if semver.Compare(kversion, "4.14") >= 0 {
+            return true, nil
+        }
+	}
+	return false, nil;
+}
+
+func getHostSupportedLsm() string {
+
+    b, err := ioutil.ReadFile("/sys/kernel/security/lsm")
```
daemon1024, line 64: ioutil was deprecated in 1.16, Let's avoid it.
Ref https://go.dev/doc/go1.16#ioutil

probe/probe.go
```diff
@@ -0,0 +1,87 @@
...
+    "io/ioutil"
+    "log"
+
+)
+
+func PrintProbeResult() error{
+    
+    fmt.Printf("Host: \n")
+    fmt.Printf("\t Observability/Audit: %s \n", checkHostAuditSupport())
+    fmt.Printf("\t Enforcement: %s \n", checkHostLsmSupport())
```
daemon1024, line 21: Let's prettify the print, WDYT?
Like showing supported in Green, Partial in Orange maybe. Bold for the Host or something.

Perhaps can leverage https://github.com/fatih/color or some more suiting lib.

probe/probe.go
```diff
@@ -0,0 +1,87 @@
...
+	return nil
+}
+
+func checkHostAuditSupport() string {
+    hostKernelVersion, err := detectHostKernelVersion()
+	if err != nil {
+		return ""
```
daemon1024, line 28:
```suggestion
		return "Error"
```
to let the users know that we had an Error while trying to Probe

daemon1024 (changes requested):
Left some comments inline.

Can you please check and fix the Linter Warnings?
And there's an empty file `running.config` committed for some reason, can that be reverted?

Thank You :)

probe/probe.go
Essietom: well noted

Essietom:
issues has been resolved, kindly review

probe/probe.go
```diff
@@ -0,0 +1,122 @@
...
+	
+}
+
+func checkNodeAuditSupport(c *k8s.Client) {
+    whilte := color.New(color.FgWhite)
+    boldWhite := whilte.Add(color.Bold)
+	nodes, _ := c.K8sClientset.CoreV1().Nodes().List(context.Background(), metav1.ListOptions{})
+     n := 0
+     if len(nodes.Items) > 0 {
+        for(n < len(nodes.Items)){
```
daemon1024, line 102: Let's perhaps use https://go.dev/tour/moretypes/16

We can eliminate the need to increment the iterator

probe/probe.go
```diff
@@ -0,0 +1,122 @@
...
+        
+	}else{
+        color.Red(" Error")
+    }
+	
+}
+
+func checkNodeAuditSupport(c *k8s.Client) {
+    whilte := color.New(color.FgWhite)
+    boldWhite := whilte.Add(color.Bold)
```
daemon1024, line 98: We could declare these outside of function, globally in this package, since being used by multiple function.

daemon1024 (changes requested):
Nice progress 🚀 
Can you share how's the output screen is looking like? 

Left some nits inline.
Also can you take a look at the failing CI?

Thank You 😸

probe/probe.go
```diff
@@ -0,0 +1,122 @@
...
+    "github.com/kubearmor/kubearmor-client/k8s"
+	"context"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func PrintProbeResult(c *k8s.Client) error{
+    
+    checkHostAuditSupport()
+    checkLsmSupport(getHostSupportedLsm())
```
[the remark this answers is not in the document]
Essietom: Screenshot 2022-06-20 at 10 16 07
sample output

Essietom:
Screenshot 2022-06-24 at 10 21 45
This is the output

nyrahul:
@Essietom , can you rebase the code to latest and squash the commits?

deployment/probeinstaller.go
```diff
@@ -0,0 +1,67 @@
...
+        daemonset.Spec.Template.Spec.Containers[0].Args = append(daemonset.Spec.Template.Spec.Containers[0].Args, "-defaultFilePosture=audit")
+    }
+    if o.Audit == "all" || strings.Contains(o.Audit, "network") {
+        daemonset.Spec.Template.Spec.Containers[0].Args = append(daemonset.Spec.Template.Spec.Containers[0].Args, "-defaultNetworkPosture=audit")
+    }
+    if o.Audit == "all" || strings.Contains(o.Audit, "capabilities") {
+        daemonset.Spec.Template.Spec.Containers[0].Args = append(daemonset.Spec.Template.Spec.Containers[0].Args, "-defaultCapabilitiesPosture=audit")
+    }
+    // fmt.Printf("Probe DaemonSet %s %v...\n", daemonset.Spec.Template.Spec.Containers[0].Image, daemonset.Spec.Template.Spec.Containers[0].Args)
+
```
daemon1024, line 42: Why do we need these here 😅 ?

I believe these are specific to KubeArmor. We shouldn't be needing to detect an environment since Probe will be the one detecting environment for us or modifying the Image and Args...

cmd/probe.go
```diff
@@ -0,0 +1,62 @@
...
+
+var probeInstallOptions deployment.ProbeOptions
+
+// probeCmd represents the get command
+var probeCmd = &cobra.Command{
+	Use:   "probe",
+	Short: "Display probe information",
+	Long:  `Display probe information`,
+	RunE: func(cmd *cobra.Command, args []string) error {
+		color.Blue("To get full probe, a pod will be deployed in your cluster - This pod will be deleted after probing")
```
daemon1024, line 24: I was hoping for an Initial probe result, before we proceed with a full probe.
```suggestion
		color.Blue("To get full probe, a daemonset will be deployed in your cluster - This daemonset will be deleted after probing")
```

cmd/probe.go
```diff
@@ -0,0 +1,62 @@
...
+		
+		
+	},
+}
+
+func init() {
+	rootCmd.AddCommand(probeCmd)
+
+	probeCmd.Flags().StringVarP(&probeInstallOptions.Namespace, "namespace", "n", "default", "Namespace for resources")
+    probeCmd.Flags().StringVarP(&probeInstallOptions.ProbeDaemonImage, "image", "i", "alpine", "Probe daemonset image to use")
```
daemon1024, line 61: We won't need to change the image here I believe.
But We should add an option to bypass the probe interrupt. Like a `-f|--full` or `-y|--yes` flag which will automatically do a full probing.

daemon1024 (changes requested):
Nice Work 🚀 

Some comments inline.

Can you handle the linter errors and squash the commits?

Essietom:
Screenshot 2022-07-04 at 07 08 55
New output

cmd/probe.go
```diff
@@ -0,0 +1,32 @@
...
+		}
+		return nil
+		
+	},
+}
+
+func init() {
+	rootCmd.AddCommand(probeCmd)
+	probeCmd.Flags().StringVarP(&probeInstallOptions.Namespace, "namespace", "n", "default", "Namespace for resources")
+	probeCmd.Flags().StringVarP(&probeInstallOptions.ProbeDaemonImage, "image", "i", "alpine", "Kubearmor daemonset image to use")
```
daemon1024, line 30:
```suggestion
	probeCmd.Flags().StringVarP(&probeInstallOptions.ProbeDaemonImage, "image", "i", "alpine", "Probe Daemonset Image to use")
```

What would be the usecase here for providing custom image?

deployment/probedeployment.go
```diff
@@ -0,0 +1,114 @@
...
+    corev1 "k8s.io/api/core/v1"
+)
+
+var karmorprobe = "karmor-probe"
+
+// GenerateDaemonSet Function
+func GenerateDaemonSet(namespace string) *appsv1.DaemonSet {
+
+    var label = map[string]string{
+        "k8s-app": karmorprobe,
```
daemon1024, line 15:
```suggestion
        "kubearmor-app": karmorprobe,
```

probe/probe.go
```diff
@@ -0,0 +1,315 @@
...
+	"k8s.io/client-go/tools/remotecommand"
+	"context"
+	corev1 "k8s.io/api/core/v1"
+	"github.com/kubearmor/kubearmor-client/deployment"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+var white = color.New(color.FgWhite)
+var boldWhite = white.Add(color.Bold)
+var karmorprobe = "karmor-probe"
```
daemon1024, line 28: Let's import the name declared in deployment package? WDYT? Could avoid duplication if we need to change names sometime.

probe/probe.go
```diff
@@ -0,0 +1,315 @@
...
+    }
+    return nil
+
+
+}
+
+
+func probeDaemonUninstaller(c *k8s.Client, o ProbeOptions) error {
+    color.Yellow("\t Deleting Karmor Probe DaemonSet ...\n")
+	if err := c.K8sClientset.AppsV1().DaemonSets(o.Namespace).Delete(context.Background(), karmorprobe, metav1.DeleteOptions{}); err != nil {
```
daemon1024, line 57:
```suggestion
	if err := c.K8sClientset.AppsV1().DaemonSets(o.Namespace).Delete(context.Background(), deployment.KarmorProbe, metav1.DeleteOptions{}); err != nil {
```

probe/probe.go
```diff
@@ -0,0 +1,315 @@
...
+    }
+    return nil
+}
+
+
+func checkLsmSupport(supportedLSM string) {
+    fmt.Printf("\t Enforcement:") 
+    if strings.Contains(supportedLSM, "selinux"){
+        color.Yellow(" Partial (Supported LSMs: "+ supportedLSM + ") \n\t To have full enforcement support, apparmor must be supported")
+    }else if strings.Contains(supportedLSM, "apparmor"){
```
daemon1024, line 96:
```suggestion
    }else if strings.Contains(supportedLSM, "apparmor") || strings.Contains(supportedLSM, "bpf") {
```

Since we recently added support for BPF LSM as part of v0.5 release.

probe/probe.go
```diff
@@ -0,0 +1,315 @@
...
+        color.Green(" Full (Supported LSMs: "+ supportedLSM + ")")
+    }else{
+        color.Red(" None (Supported LSMs: "+ supportedLSM + ") \n\t To have full enforcement support, apparmor must be supported")
```
daemon1024, line 99:
```suggestion
        color.Red(" None (Supported LSMs: "+ supportedLSM + ") \n\t To have full enforcement support, AppArmor or BPFLSM must be supported")
```

probe/probe.go
```diff
@@ -0,0 +1,315 @@
...
+    }
+}
+
+func getHostSupportedLSM() string {
+
```
daemon1024, line 104:
```suggestion
```

probe/probe.go
```diff
@@ -0,0 +1,315 @@
...
+        var path = ""
+        if _, err := os.Stat("/etc/redhat-release"); !os.IsNotExist(err) {
+            path = "/usr/src/"+int8ToStr(uname.Release[:])
+        }else{
+            path = "/usr/src/linux-headers-"+int8ToStr(uname.Release[:])
+        }
+       
+        if _, err := os.Stat(path); !os.IsNotExist(err) {
+            return true
+        }
```
daemon1024, line 149: I have updated the Kernel Header logic check here
https://github.com/kubearmor/KubeArmor/blob/main/KubeArmor/BPF/Makefile#L3-L11

Can you see if we can handle similarly here?
We meet need to add an additional mount point for "/lib/modules"

probe/probe.go
```diff
@@ -0,0 +1,315 @@
...
+        FieldSelector: "spec.nodeName=" + nodeName,
+	})
+	if err != nil {
+		return false
+	}
+    srcPath :=  "/usr/src/"
+    fileName := "*"+kernelVersion+"*"
+    reader, outStream := io.Pipe()
+
+    cmdArr := []string{"find", srcPath, "-name", fileName}
```
daemon1024, line 168: Wrt above review comment, will need to handle lib modules check here as well ig.

probe/probe.go
```diff
@@ -0,0 +1,315 @@
...
+        fmt.Printf("\t Observability/Audit:") 
+        checkAuditSupport(kernelVersion, checkKernelHeaderPresent())
+        
+	}else{
+        color.Red(" Error")
+    }
+	
+}
+
+
```
daemon1024, line 244:
```suggestion

```

probe/probe.go
```diff
@@ -0,0 +1,315 @@
...
+		}
+		fmt.Print("Karmor probe DaemonSet not found ...\n")
+	}
+
+
+	return nil
+}
+
+func PrintProbeResult(c *k8s.Client, o ProbeOptions) error{
+    if o.Full {
```
daemon1024, line 69:
```suggestion
	print("Didn't find KubeArmor in systemd or Kubernetes, probing for support for KubeArmor
    if o.Full {
```
This might be handled in the following PR for probing with KubeArmor present?

daemon1024 (changes requested):
🚀 

Find some comments, feedback and nitpicks inline 😁

cmd/probe.go
Essietom: No use case actually, forgot to take that out

probe/probe.go
Essietom: > Wrt above review comment, will need to handle lib modules check here as well ig.

pending

probe/probe.go
```diff
@@ -0,0 +1,315 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2021 Authors of KubeArmor
+
+package probe
+
+
+import (
```
daemon1024, line 7:
```suggestion
// SPDX-License-Identifier: Apache-2.0
// Copyright 2022 Authors of KubeArmor

package probe

import (
```

probe/probe.go
```diff
@@ -0,0 +1,315 @@
...
+	"github.com/kubearmor/kubearmor-client/deployment"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+var white = color.New(color.FgWhite)
+var boldWhite = white.Add(color.Bold)
+
+
+
+// Options for probe daemonset options install
```
daemon1024, line 31:
```suggestion
var boldWhite = white.Add(color.Bold)

// Options for probe daemonset options install
```

probe/probe.go
```diff
@@ -0,0 +1,315 @@
...
+            return err
+        }
+    }
+    return nil
+
+
+}
+
+
+func probeDaemonUninstaller(c *k8s.Client, o ProbeOptions) error {
```
daemon1024, line 52:
```suggestion
    return nil
}

func probeDaemonUninstaller(c *k8s.Client, o ProbeOptions) error {
```

probe/probe.go
```diff
@@ -0,0 +1,315 @@
...
+    color.Yellow("\t Deleting Karmor Probe DaemonSet ...\n")
+	if err := c.K8sClientset.AppsV1().DaemonSets(o.Namespace).Delete(context.Background(), deployment.Karmorprobe, metav1.DeleteOptions{}); err != nil {
+		if !strings.Contains(err.Error(), "not found") {
+			return err
+		}
+		fmt.Print("Karmor probe DaemonSet not found ...\n")
+	}
+
+
+	return nil
```
daemon1024, line 62:
```suggestion

	return nil
```

probe/probe.go
```diff
@@ -0,0 +1,315 @@
...
+        if err := probeDaemonInstaller(c, o); err != nil {
+            return err
+        }
+        color.Yellow("\t Creating probe daemonset ...")
+        time.Sleep(2 * time.Second)
+        probeNode(c)
+        if err := probeDaemonUninstaller(c, o); err != nil {
+            return err
+        }
+	}else{
```
daemon1024, line 79:
```suggestion
	} else {
```

probe/probe.go
daemon1024: Can we handle this across the package.

probe/probe.go
```diff
@@ -0,0 +1,315 @@
...
+
+func checkAuditSupport(kernelVersion string, kernelHeaderPresent bool) {
+
+    if kernelVersionSupported(kernelVersion) && kernelHeaderPresent{
+        color.Green(" Supported (Kernel Version " + kernelVersion +")")
+    }else if(kernelVersionSupported(kernelVersion)){
+        color.Red(" Not Supported : Kernel header must be present")
+    }else{
+        color.Red(" Not Supported (Kernel Version " + kernelVersion + " \n\t Kernel version must be greater than 4.14) and kernel header must be present")
+    }
```
daemon1024, line 125:
```suggestion
    if kernelVersionSupported(kernelVersion) && kernelHeaderPresent {
        color.Green(" Supported (Kernel Version " + kernelVersion +")")
    } else if (kernelVersionSupported(kernelVersion)){
        color.Red(" Not Supported : Kernel header must be present")
    } else {
        color.Red(" Not Supported (Kernel Version " + kernelVersion + " \n\t Kernel version must be greater than 4.14) and kernel header must be present")
    }
```

probe/probe.go
```diff
@@ -0,0 +1,315 @@
...
+            path = "/usr/src/linux-headers-"+int8ToStr(uname.Release[:])
+        }
+       
+        if _, err := os.Stat(path); !os.IsNotExist(err) {
+            return true
+        }        
+        
+	}
+     return false
+}
```
daemon1024, line 152:
```suggestion
	}
     return false
}
```

probe/probe.go
```diff
@@ -0,0 +1,315 @@
...
+	buf, err := io.ReadAll(reader)
+	if err != nil {
+		return false
+	}
+    s := string(buf)
+    if strings.Contains(s, "No such file or directory") || len(s) == 0{
+        return false
+    }else{
+        return true
+    }
```
daemon1024, line 205:
```suggestion
	}
	
	go func() {
		defer outStream.Close()
		err = exec.Stream(remotecommand.StreamOptions{
			Stdin:  os.Stdin,
			Stdout: outStream,
			Stderr: os.Stderr,
			Tty:    false,
		})
	}()
	
	buf, err := io.ReadAll(reader)
	if err != nil {
		return false
	}
	
    s := string(buf)
    if strings.Contains(s, "No such file or directory") || len(s) == 0 {
        return false
    } else {
        return true
    }
```

probe/probe.go
```diff
@@ -0,0 +1,315 @@
...
+    
+}
+
+
+
+ // A utility to convert the values to proper strings.
+func int8ToStr(arr []int8) string {
```
daemon1024, line 212:
```suggestion
}

 // A utility to convert the values to proper strings.
func int8ToStr(arr []int8) string {
```

probe/probe.go
```diff
@@ -0,0 +1,315 @@
...
+            checkLsmSupport(lsm)
+            
+        }
+    }else{
+        fmt.Println("No kubernetes environment found")
+     }
+
+}
+
+
```
daemon1024, line 315:
```suggestion
            checkLsmSupport(lsm)        
        }
    } else {
        fmt.Println("No kubernetes environment found")
     }
}
```

daemon1024 (changes requested):
Can you handle the formatting?

daemon1024 (approved):

Issue #79, opened 2022-06-09
kubearmor policy simulator


nyrahul:
## What is the problem statement and use-case for this?
Kubearmor supports policy enforcement and this enforcement is guided by two factors:
1. The policy itself
2. Kubearmor configuration (such as `defaultPosture`, `auditOnlyMode` etc)

It is possible to put up a document to explain all the combination as in, if X is the policy and Y is the configuration and if we get an action exec/fopen/network-connect etc, how will the policy react? However, using a document for such purpose seems non-efficient. Also it can very easily get complicated since every configuration has multiple values (for e.g. `defaultPosture` could be Audit/Block/Allow). As an example, check the discussion on [this PR](https://github.com/kubearmor/KubeArmor/pull/727). It  is better if we can simulate the event/action-result given the inputs. Later on, we can have a web-page hosted that can depict this in a more user friendly way.

## Example cli
```
karmor simulate --config kubearmor.cfg --policy policy.yaml --action "exec:/bin/sleep"
karmor simulate --config kubearmor.cfg --policy policy.yaml --action "exec:/bin/bash->/bin/sleep" ... specifying ls spawned as a child process of bash
```
<details>
  <summary>Sample policy.yaml</summary>
  
```yaml
apiVersion: security.kubearmor.com/v1
kind: KubeArmorPolicy
metadata:
  name: ksp-group-1-proc-path-block
  namespace: multiubuntu
spec:
  selector:
    matchLabels:
      group: group-1
  process:
    matchPaths:
    - path: /bin/sleep
  action:
    Block
```
</details>

### Sample run
Assuming the command used is:
```
karmor simulate --config kubearmor.cfg --policy policy.yaml --action "exec:/bin/sleep"
```

Expected Output:
```
Action: Block

Telemetry Event:
== Alert ==
Cluster Name: unknown
Host Name: unknown
Namespace Name: unknown
Pod Name: unknown
Container ID: unknown
Container Name: unknown
Labels: unknown
Policy: policy.yaml
Severity: 1
Type: MatchedPolicy
Source: /bin/sleep
Operation: Process
Resource: /bin/sleep
Data: syscall=SYS_EXECVE
Action: Block
Result: Permission denied
```

## Requirements:
- [ ] support multiple policies
- [ ] support configuration file for specifying kubearmor cfg
- [ ] support process based rules
- [ ] file based rules
- [ ] network based rules
- [ ] show the output action
- [ ] show the telemetry event that would be generated
[...]

PR #87, opened 2022-06-15
Policy Simulation command [WIP]

s1ntaxe770r:
WIP implementation of #79

### DIFF
--- a/simulate/simulate.go
+++ b/simulate/simulate.go
@@ -16,6 +16,7 @@ type Options struct {
 	Action string
 	Policy string
 }
+
 type SimulationOutput struct {
 	Policy    string
 	Severity  int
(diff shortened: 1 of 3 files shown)


nyrahul:
> @nyrahul this is far from done here but what do you think of this current implementation? Also, what does `kubearmor.cfg` look like ?

### kubearmor.cfg
```
defaultPosture: audit # possible values allow, block, audit
```
This is the only configuration that needs to be used as of now since this is required for policy simulation

s1ntaxe770r:
Thanks, also I see over [here](https://github.com/kubearmor/KubeArmor/pull/727) that default posture is passed as an environment variable, Should the command take this into account? If yes should it take preference if a `kubearmor.cfg` is passed in

simulate/simulate.go
```diff
@@ -0,0 +1,116 @@
...
+		response.Action = karmorPolicy.Spec.Process.Action
+		response.Result = "Permission Denied"
+		PrintResults(response, "Block")
+		return nil
+	} else if len(karmorPolicy.Spec.File.MatchPaths) > 0 {
+		response.Resource = karmorPolicy.Spec.File.MatchPaths[0].Path
+		response.Policy = o.Policy
+		response.Severity = 1
+		response.Source = karmorPolicy.Spec.File.MatchPaths[0].Path
+		response.Data = "syscall=SYS_FOPEN"
```
s1ntaxe770r, line 78: hey, @nyrahul  is this the right syscall?

nyrahul: Nope, you can setup kubearmor and check the telemetry events for file open.

s1ntaxe770r:
@nyrahul here's how it's looking right now, still need to handle more conditions but i hope i am on the right track here
image

Ankurk99:
Hey @s1ntaxe770r are you working on this issue?

s1ntaxe770r:
Right now no , I got stuck on some on parts of the design spec so I'm working on other issues right now

PR #93, opened 2022-06-25
Add support for crio [Vikas Verma]

vikasvr:
Referencing issue: https://github.com/kubearmor/KubeArmor/issues/221

Added support for cri-o.

Before: Unable to run `karmor install` on openshift due to missing runtime support for `cri-o`
Now: I was able to run `karmor install` successfully without any errors.

Signed-off-by: Vikas Verma <vikasvr23@gmail.com>

### DIFF
--- a/install/install.go
+++ b/install/install.go
@@ -399,7 +399,7 @@ func autoDetectEnvironment(c *k8s.Client) (name string) {
 		env = "docker"
 		return env
 	}
-	if (runtime == "docker" && semver.Compare(version, "v19.3") >= 0) || runtime == "containerd" {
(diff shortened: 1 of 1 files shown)


daemon1024:
Nice!

> Now: I was able to run karmor install successfully without any errors.

`karmor install` would work successfully with this but the installed daemonset wouldn't work as intended, since it would have mount paths based on generic env i.e. containerd sockets instead of crio ones.

We recently added support for CRI-O in KubeArmor as part of https://github.com/kubearmor/KubeArmor/issues/666.

Since we can already detect crio here, let's add support for new env.
This would require modifications in our [`deployments/get` pkg](https://github.com/kubearmor/KubeArmor/tree/main/deployments/get) which can based off of https://github.com/kubearmor/KubeArmor/blob/main/KubeArmor/build/kubearmor-test-crio.yaml#L80=L85

PR #106, opened 2022-07-19
Observe cmd init

nthnieljson:


observe/observe.go
```diff
@@ -0,0 +1,144 @@
...
+	Operation     string
+	Namespace     string
+	AllNamespace  bool
+	Labels        string
+	ShowLabels    bool
+	Since         string
+	CustomColumns string
+}
+
+// Get observability data
```
seswarrajan, line 30: Fix linter error

observe/observe.go
```diff
@@ -0,0 +1,144 @@
...
+	proto "github.com/kubearmor/koach/protobuf"
+	"github.com/rodaine/table"
+	"google.golang.org/grpc"
+)
+
+// Options Structure
+type Options struct {
+	Operation     string
+	Namespace     string
+	AllNamespace  bool
```
seswarrajan, line 23: If above option(Namespace) is null/empty, can we cover all namespaces ?

seswarrajan (changes requested):

observe/observe.go
nthnieljson: no.

the behaviour is the same kubectl, if it is empty, then we will assume it is "default" namespace, I have provided "-A" flag to cover all namespaces

Ankurk99:
Hey @nthnieljson! Can you please handle this PR?

daemon1024 (approved):

daemon1024:
Ref https://github.com/kubearmor/koach/pull/7

yasin-cs-ko-ak:
@nthnieljson Kindly resolve this conflict and also handle `go-sec` CI.
cc: @daemon1024 @seswarrajan

PR #107, opened 2022-07-21
Handling removal of pod annotations

seswarrajan:


seswarrajan:
The issue is handled in annotation controller

Issue #112, opened 2022-08-08
policy recommendations/reporting


nyrahul:
## Aim
1. identify possible security gaps
1. recommend policies based on
     1.  container image
     2. k8s deployment manifest
     3. runtime data
1. keep performance impact in mind

## Based on container image
```
karmor report --image "homeassistant:latest" --yamldir "policies" --output report.pdf
    ... yamldir will contain the set of recommended policies
```

1. Audit /sbin/ ... reason
3. disable write to /boot folder on the host
2.     Block access to following folders recursively:
```
        /usr/share/ca-certificates
        /etc/ssl/
        recursive: true
```
Block process execution
```
/usr/sbin/update-ca-certificates
```

namespace
    deployment/workload
        Application
            1. Description: audit access to `/sbin/` 
               Reason: sbin contains maintainance tools .... (tooltip)
               ...
            2. 
  

## Based on k8s manifest
```
karmor report --k8s-manifest deployment.yaml
```
1. based on container images ... recommend polices
2. based on volume mount points ... 
5. k8s secrets ... audit k8s secrets

## Based on runtime behavior
```
karmor report -n namespace -l "app=wordpress"
```
1. based on container images ... recommend polices
2. based on volume mount points ... 
4. k8s secrets ... we see these bins accessing these k8s-secrets ... 
6. Processes -> 80% of bins never used ... 10% are risky .. audit policy recommended

## Based on kube-bench/trivy
<todo>future</todo>
1. mapping it to compliance, mitre TTPs
2. STIG hardening 

## Output 

json => html/markdown/stdout => pdf
Look at the output of https://github.com/aquasecurity/trivy

### design 

[How do we want to structure the final report?]

[common way to report an output]
Reporting API

[image scanner]
[k8s scanner]
[runtime scanner]

nyrahul:
# Handling Policy Recommendations

## container-img
* keep rules in yaml format
* processRule to be added to rules.json
* remove empty rules `network: {}` 
* networkRule to be added to rules.json
* owneronly flag (Rahul)

## policy-templates
- [ ] update rules.json from policy-templates

## k8s-manifest
* deployment container images
* identifying namespace and labels
* protecting k8s service account 
* protecting k8s secrets ... /etc/xyz/ => secret_xyz

# runtime (based on discovery-engine)
* k8s service account needs to be enabled access for specific process
* k8s secrets needs to be enabled access for specific process

nyrahul:
[...]

PR #114, opened 2022-08-09
policy recommend option for karmor

nyrahul:
Ref: #112 

Signed-off-by: Rahul Jadhav <nyrahul@gmail.com>


nyrahul:
Sample Reports:
1. [report.zip](https://github.com/kubearmor/kubearmor-client/files/9334100/report.zip) (html report... compressed to zip since GH does not allow html to be attached)
2. [report.txt](https://github.com/kubearmor/kubearmor-client/files/9334101/report.txt)

recommend/imageHandler.go
```diff
@@ -0,0 +1,362 @@
...
+	tarname := saveImageToTar(imageName)
+
+	// step 2: retrieve information from tar
+	img.FileList, img.DirList = extractTar(tarname)
+
+	// step 3: getImageInfo
+	getImageInfo(&img)
+
+	// step 4: get policy from image info
+	getPolicyFromImageInfo(&img)
```
daemon1024, line 343: Essentially we are reading the tar file and appending individual file/dir to arrays.
And then doing a regex match on each element in the array using based on PreCondition?

Is that right?

recommend/policyRules.go
```diff
@@ -0,0 +1,80 @@
...
+	Severity int      `json:"severity"`
+	Message  string   `json:"message"`
+	Tags     []string `json:"tags"`
+	Action   string   `json:"action"`
+}
+
+var policyMatch []MatchSpec
+
+//go:embed json/rules.json
+var ruleSpecJSON []byte
```
daemon1024, line 60: I am not sure if we should be embedding rules.json. The update frequency for this rule list will be pretty high and it wouldn't be feasible to release a new binary for each update to rules. We should perhaps host it at policy templates maybe and then  do a get request?

daemon1024 (changes requested):
This is really cool 🤩🤩🤩

Find some comments inline.

recommend/imageHandler.go
```diff
@@ -0,0 +1,362 @@
...
+	matches := checkForSpec(filepath.Join(tempDir, "manifest.json"), img.FileList)
+	if len(matches) != 1 {
+		log.WithFields(log.Fields{
+			"len":     len(matches),
+			"matches": matches,
+		}).Fatal("expecting one manifest.json!")
+	}
+	readManifest(img, matches[0])
+
+	getDistro(img)
```
daemon1024, line 327:
```suggestion
func (img *ImageInfo) getImageInfo() {
	matches := checkForSpec(filepath.Join(tempDir, "manifest.json"), img.FileList)
	if len(matches) != 1 {
		log.WithFields(log.Fields{
			"len":     len(matches),
			"matches": matches,
		}).Fatal("expecting one manifest.json!")
	}
	readManifest(img, matches[0])

	img.getDistro()
```

What do you think about implementing these helper functions related to image as struct methods, instead of passing image address each time.
Other instances can be
- img.extractTar()
- img.getPolicy()

Makefile
```diff
@@ -13,7 +13,7 @@ GIT_INFO := $(shell govvv -flags -pkg $(PKG))
 
 .PHONY: build
 build:
-	cd $(CURDIR); go mod tidy; CGO_ENABLED=0 go build -ldflags "-w -s ${GIT_INFO}" -o karmor
+	cd $(CURDIR); go mod tidy; go build -ldflags "-w -s ${GIT_INFO}" -o karmor
```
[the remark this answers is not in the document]
nyrahul: We dont need to enable CGO. I enabled it to test something and forgot to disable it. Thanks for pointing out.

recommend/policy.go
```diff
@@ -0,0 +1,172 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2022 Authors of KubeArmor
+
+package recommend
+
+import (
+	"encoding/json"
```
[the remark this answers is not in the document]
nyrahul: super. It took care of empty constructs 👍🏼

go.mod
```diff
@@ -11,24 +11,34 @@ replace (
 )
 
 require (
+	github.com/Microsoft/go-winio v0.5.2 // indirect
```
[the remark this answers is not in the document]
nyrahul: done

README.md
```diff
@@ -13,6 +13,7 @@ curl -sfL http://get.kubearmor.io/ | sudo sh -s -- -b /usr/local/bin
 Build karmor from source if you want to test the latest (pre-release) karmor version.
 
 ```
+sudo apt-get install libbtrfs-dev -y
```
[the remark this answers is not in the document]
nyrahul: removing this. Not needed on client-side. Great catch.

recommend/imageHandler.go
nyrahul: Yep, that's right

recommend/policyRules.go
nyrahul: We will pull the rules from the policy-template eventually like we had discussed. We do not want to keep a separate configuration file hence embedding the rules as part of the bin itself.

We will have two things done next (in subsequent PR):
1. have a way to pass an explicit rules.json in addition as an input from command line
2. update the rules.json from the policy-templates

recommend/imageHandler.go
nyrahul: Yes, this makes sense. Have modified for all the relevant places.

cmd/recommend.go
```diff
@@ -0,0 +1,33 @@
...
+package cmd
+
+import (
+	"github.com/kubearmor/kubearmor-client/recommend"
+	"github.com/spf13/cobra"
+)
+
+var recommendOptions recommend.Options
+
+// recommendCmd represents the get command
```
daemon1024, line 13:
```suggestion
// recommendCmd ....
```

recommend/imageHandler.go
```diff
@@ -0,0 +1,366 @@
...
+	FileList []string
+	DirList  []string
+}
+
+func getAuthStr() string {
+	u := os.Getenv("DOCKER_USERNAME")
+	p := os.Getenv("DOCKER_PASSWORD")
+	if u == "" || p == "" {
+		return ""
+	}
```
daemon1024, line 50: What do we need these for? Private images?

recommend/imageHandler.go
```diff
@@ -0,0 +1,366 @@
...
+
+func (img *ImageInfo) getImageInfo() {
+	matches := checkForSpec(filepath.Join(tempDir, "manifest.json"), img.FileList)
+	if len(matches) != 1 {
+		log.WithFields(log.Fields{
+			"len":     len(matches),
+			"matches": matches,
+		}).Fatal("expecting one manifest.json!")
+	}
+	img.readManifest(matches[0])
```
daemon1024, line 327: What do you think about using https://pkg.go.dev/github.com/containers/image/manifest to read the manifest here?

recommend/policyRules.go
```diff
@@ -0,0 +1,82 @@
...
+type PathRule struct {
+	FromSource string   `json:"fromSource"`
+	Path       []string `json:"path"`
+	Recursive  bool     `json:"recursive"`
+	Owneronly  bool     `json:"owneronly"`
+}
+
+// Rules set of applicable rules. In the future, we might have other types of rules.
+type Rules struct {
+	PathRule PathRule `json:"pathRule"`
```
daemon1024, line 48: Will we handle addition of Process/File/Network in a subsequent PR?

recommend/imageHandler.go
nyrahul: Yep, for private images. I thought it would be better to have atleast some option to specify username/token

recommend/imageHandler.go
nyrahul: Awesome find. Let me check if I can use it.

recommend/policyRules.go
nyrahul: Yes, there is already a discussion on this. We will need this immediately but we will take care in a separate PR.

recommend/imageHandler.go
nyrahul: Hey @daemon1024 , using `containers/image` would be incredible i.e., it will allow us to work with any container registry, it will provide most of the other manifest API.

One problem that I saw with `containers/image` APIs is that it requires root access since it assumes that the containers are mounted as part of `/var/lib/containers`. ([ref](https://iximiuz.com/en/posts/working-with-container-images-in-go/))

Currently I am relying on docker friendly APIs and imo for the current version it should be good enough without requiring root privileges.

daemon1024: Okay, let's keep it in mind for future then.

daemon1024 (approved):

PR #117, opened 2022-08-12
chore: fix ci, handle ci warnings and updates to install package

daemon1024:


nyrahul (approved):
deleting the CRDs only in the `--force` option? Thanks for fixing the fancy-formatting issues. LGTM 👍🏼 
Cant believe you fixed all the lint warnings too! Awesome!!

daemon1024:
> deleting the CRDs only in the `--force` option? 

This would help preserve the policies even after uninstallation 👀 WDYT?
Lingering policies won't really do anything on it's own unless we install KubeArmor back. So imo it should be fine.
This was as per the discussion we had with @achrefbensaad

PR #120, opened 2022-08-16
fix sysname error and os probe support

Essietom:
Signed-off-by: essietom <essietom@gmail.com>


install/install.go
```diff
@@ -550,7 +549,7 @@ func K8sUninstaller(c *k8s.Client, o Options) error {
 	return nil
 }
 
-func autoDetectEnvironment(c *k8s.Client) (name string) {
+func AutoDetectEnvironment(c *k8s.Client) (name string) {
```
daemon1024, line 552: Can you add a simple comment in the format of `// AutoDetectEnvironment detect the environment for a given k8s context`

probe/probe.go
```diff
@@ -57,8 +62,14 @@ func probeDaemonUninstaller(c *k8s.Client, o Options) error {
 	return nil
 }
 
-// PrintProbeResult prints the result for the  host and k8s probing kArmor does to check compatibility with KubeArmor
```
daemon1024, line 60: Can you please add this back?

daemon1024 (changes requested):

probe/probe.go
Essietom: ok

daemon1024 (approved):

PR #122, opened 2022-08-17
fix: blank policy check

slayer321:

Add check for the manifest file after the separator (---)  if there is a blank .

fixed [#769](https://github.com/kubearmor/KubeArmor/issues/769)

### DIFF
--- a/vm/policy.go
+++ b/vm/policy.go
@@ -11,6 +11,7 @@ import (
 	"net/http"
 	"os"
 	"path/filepath"
+	"regexp"
 	"strings"
 	"time"
 
@@ -121,7 +122,8 @@ func PolicyHandling(t string, path string, o PolicyOptions, httpAddress string, 
 	policies := strings.Split(string(policyFile), "---")
 
 	for _, policy := range policies {
-		if policy == "" {
+
+		if matched, _ := regexp.MatchString("^\\s*$", policy); matched {
 			continue
 		}
 


vm/policy.go
```diff
@@ -121,7 +122,10 @@ func PolicyHandling(t string, path string, o PolicyOptions, httpAddress string,
 	policies := strings.Split(string(policyFile), "---")
 
 	for _, policy := range policies {
-		if policy == "" {
+		re, _ := regexp.Compile("^\\s*$")
+		emptyPolicy := re.MatchString(policy)
+
+		if emptyPolicy {
```
nyrahul, line 128:
```suggestion
                if matched, _ := regexp.MatchString("^\\s*$", policy); matched {
```

nyrahul (changes requested):

Ankurk99:
LGTM, can you please squash your commits?

vm/policy.go
```diff
@@ -121,7 +122,8 @@ func PolicyHandling(t string, path string, o PolicyOptions, httpAddress string,
 	policies := strings.Split(string(policyFile), "---")
 
 	for _, policy := range policies {
-		if policy == "" {
+
+		if matched, _ := regexp.MatchString("^\\s*$", policy); matched {
```
daemon1024, line 126: What do you think about using https://pkg.go.dev/unicode#IsSpace instead?

slayer321: After reading the `unicode.IsSpace()` function .. I see that I takes rune as a single character . Here in the for loop we are dealing with strings. Even if we try to convert the string to rune slice we cannot pass it to the `unicode.IsSpace()`  function .
Imo regex seems fine.

daemon1024 (approved):

Ankurk99 (approved):

Issue #103, opened 2022-07-14
Add option to view list of applied policies


Ankurk99:
Add `./karmor list-policy <containername>/<podname>/host`to view list of currently applied policies.
Currently we are backing up the policies file in `/opt/kubearmor/policies`, they can be used to retrieve the applied policies.

s1ntaxe770r:
hey @Ankurk99  would like to take this one

Ankurk99:
Awesome!

nyrahul:
This change was never implemented and hence reopening.

PR #150, opened 2022-08-26
Add `list-policy` command

s1ntaxe770r:
this PR implements #103

### DIFF
--- a/list/list.go
+++ b/list/list.go
@@ -159,7 +159,7 @@ func getSystemdPolicies() ([]HostPolicyBackup, error) {
 	}
 	policies := []HostPolicyBackup{}
 	for _, file := range files {
-		path := filepath.Join("/opt/kubearmor/policies/%s", filepath.Clean(file.Name()))
+		path := filepath.Join("/opt/kubearmor/policies", filepath.Clean(file.Name()))
 		fileBytes, err := os.ReadFile(path)
 		if err != nil {
 			return nil, err
--- a/list/list.go
+++ b/list/list.go
@@ -13,9 +13,9 @@ import (
 	v1 "github.com/kubearmor/KubeArmor/pkg/KubeArmorPolicy/api/security.kubearmor.com/v1"
 	"github.com/kubearmor/kubearmor-client/k8s"
 	"github.com/olekukonko/tablewriter"
-	"gopkg.in/yaml.v2"
 	kerr "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/yaml"
 )
 
 type Options struct {
@@ -80,11 +80,7 @@ func printSystemdPolices() error {
 	policies := []tp.K8sKubeArmorHostPolicy{}
 	for _, file := range files {
 		path := fmt.Sprintf("/opt/kubearmor/policies/%s", file.Name())
-		reader, err := os.Open(path)
-		if err != nil {
-			return err
-		}
-		fileBytes, err := ioutil.ReadAll(reader)
+		fileBytes, err := os.ReadFile(path)
 		if err != nil {
 			return err
 		}
--- a/list/list.go
+++ b/list/list.go
@@ -80,12 +80,15 @@ func printSystemdPolices() error {
 	policies := []tp.K8sKubeArmorHostPolicy{}
 	for _, file := range files {
 		path := fmt.Sprintf("/opt/kubearmor/policies/%s", file.Name())
-		fileBytes, err := ioutil.ReadFile(path)
+		reader, err := os.Open(path)
+		if err != nil {
+			return err
+		}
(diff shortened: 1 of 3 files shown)


daemon1024 (commented):
I think we should implement listing/getting policies from KubeArmor in systemd/VM mode before we can merge this.

We already have an plain implementation of getting policies from K8s at https://github.com/kubearmor/kubearmor-client/blob/main/get/get.go ( This was the first command, how I started the project :P )

s1ntaxe770r:
oh yeah I'm yet to finish that. Working on it

daemon1024:
Ooof, Do you think it's okay to put in Draft mode then? That way when you convert it back we know it's ready?

Ankurk99:
@s1ntaxe770r Have you tested this in systemd mode? I am not able to run it. It's showing error: 
image

s1ntaxe770r:
Just realized I had `kubectl proxy` running when I tested this. I need to fix that

Ankurk99:
I am still getting an empty policy table.

s1ntaxe770r:
odd works for me 
image

s1ntaxe770r:
well doesn't work on my ubuntu box now. Trying to fix

s1ntaxe770r:
I figured it out. The formatting of the default yaml present in `/opt/kubearmor/policies` isn't correct so the umarshal doesn't fail but the fields simply do not match, I replaced the file with [this](https://github.com/kubearmor/KubeArmor/blob/main/examples/host-security-policies/hsp-kubearmor-dev-file-dir-allow-fromSource.yaml) and it works 

image

Ankurk99:
> The formatting of the default yaml present in /opt/kubearmor/policies isn't correct 
what do you mean by default yaml in the folder?

Ankurk99 (changes requested):
It's still not working for me. Can you please fix this and I have a few suggestions:
1) Can you also add the pod/container/host name and namespace to which the policies are applied? 
2) PTAL at the go-sec warning
Please re-request a review once you find these to be addressed.

Ankurk99:
I have applied these policies: https://raw.githubusercontent.com/kubearmor/KubeArmor/main/examples/host-security-policies/hsp-kubearmor-dev-proc-path-block.yaml, https://raw.githubusercontent.com/kubearmor/KubeArmor/main/examples/host-security-policies/hsp-kubearmor-dev-file-dir-allow-fromSource.yaml

s1ntaxe770r:
@Ankurk99 this should work now

Ankurk99:
@s1ntaxe770r Yes it's working. Can you please also address the points mentioned [here](https://github.com/kubearmor/kubearmor-client/pull/150#pullrequestreview-1090031837)?

s1ntaxe770r:
@Ankurk99 pod names and namespaces are included in the output , working on the gosec error 
image

Ankurk99:
I still couldn't see this :confused: 
image

s1ntaxe770r:
@Ankurk99  have you pulled the latest changes ?

Ankurk99:
Yes I did.

s1ntaxe770r:
oh wait ,  did you pass the `--namespace` flag? by default `list-policy` would look for policies in `kube-system`. 🤔  should I make namespace a required flag ? @Ankurk99

Ankurk99:
I think, we should be considering all the namespaces for displaying policies. `karmor list-policy` should show all the applied policies in all namespaces.

list/list.go
```diff
@@ -0,0 +1,203 @@
...
+		}
+		return []HostPolicyBackup{}, err
+	}
+	if len(files) <= 0 {
+		fmt.Printf("Unable find host polices")
+		return nil, err
+	}
+	policies := []HostPolicyBackup{}
+	for _, file := range files {
+		path := filepath.Join("/opt/kubearmor/policies", filepath.Clean(file.Name()))
```
s1ntaxe770r, line 169: @Ankurk99  the go sec error seems to be caused by this line , however I am already cleaning the file paths

s1ntaxe770r:
@Ankurk99  this should work for all namespaces now

list/list.go
Ankurk99: I think it's because we are trying to access the path which otherwise would not be possible by a non-root user.

Ankurk99:
> @Ankurk99 this should work for all namespaces now
Still can't find this working. :/
image

Hey @daemon1024 can you also pls try this out?

list/list.go
s1ntaxe770r: oh ,  I guess this is ok? wdyt?

yasin-cs-ko-ak:
As of now it is not necessary to have this command. Because we already have `karmor probe` command and it'll show all applied policies and information. We can close this PR.
Any thoughts?
cc: @Ankurk99 @daemon1024 @nyrahul

Ankurk99:
Closing as k8s based policies already handled in `karmor probe`. We can update the `karmor probe` displaying host based policies.

PR #152, opened 2022-08-29
Update karmor recommend command

vishnusomank:
- Added condition to check if at least an image is passed as an argument. 
- If no images are passed `karmor recommend -h` is shown in the output

Addresses https://github.com/kubearmor/kubearmor-client/issues/112#issuecomment-1216476912

### DIFF
--- a/cmd/recommend.go
+++ b/cmd/recommend.go
@@ -4,6 +4,8 @@
 package cmd
 
 import (
+	"errors"
+
 	"github.com/kubearmor/kubearmor-client/recommend"
 	"github.com/spf13/cobra"
 )
@@ -16,6 +18,10 @@ var recommendCmd = &cobra.Command{
 	Short: "Recommend Policies",
 	Long:  `Recommend policies based on container image, k8s manifest or the actual runtime env`,
 	RunE: func(cmd *cobra.Command, args []string) error {
+		//Condition to check if at least one Container image name is passes as an argument
+		if len(recommendOptions.Images) < 1 {
+			return errors.New("at least one container image is required as an argument")
+		}
 		if err := recommend.Recommend(client, recommendOptions); err != nil {
 			return err
 		}


cmd/recommend.go
```diff
@@ -16,6 +18,12 @@ var recommendCmd = &cobra.Command{
 	Short: "Recommend Policies",
 	Long:  `Recommend policies based on container image, k8s manifest or the actual runtime env`,
 	RunE: func(cmd *cobra.Command, args []string) error {
+		if len(recommendOptions.Images) < 1 {
+			if err := cmd.Help(); err != nil {
+				return err
+			}
+			os.Exit(0)
```
nyrahul, line 25: Hey @vishnusomank , will this show an error saying atleast one container image needs to be passed as an input?

nyrahul (changes requested):

cmd/recommend.go
vishnusomank: Hi @nyrahul it will not show an error message but instead shows the basic help for `karmor recommend`
I can add the error message if needed

vishnusomank: Added a proper error message and removed the help menu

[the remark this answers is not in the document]
vishnusomank: updated the condition with return statement

cmd/recommend.go
```diff
@@ -16,6 +18,16 @@ var recommendCmd = &cobra.Command{
...
 	RunE: func(cmd *cobra.Command, args []string) error {
+		//Condition to check if at least one Container image name is passes as an argument
+		if len(recommendOptions.Images) < 1 {
+			fmt.Printf("Error: At least one Container image is required as an argument.\n")
+			/*
+				if err := cmd.Help(); err != nil {
+					return err
+				}
+			*/
+			return nil
```
daemon1024, line 29:
```suggestion
		if len(recommendOptions.Images) < 1 {
			return errors.New("at least one container image is required as an argument.")
```

This way we preserve an exitcode which signifies an error, keeping it POSIX compliant.

daemon1024 (approved):

Ankurk99:
@vishnusomank PR looks good. Please checkout one minor go-lint warning.

vishnusomank:
Hi, @Ankurk99 Thanks for the review. I've updated the code to eliminate the lint warning

Ankurk99 (approved):
LGTM

PR #154, opened 2022-08-30
fix formatting error in karmor probe

Essietom:
Signed-off-by: Esther Adenekan <essietom@gmail.com>

### DIFF
--- a/probe/probe.go
+++ b/probe/probe.go
@@ -606,6 +606,8 @@ func getAnnotatedPods(c *k8s.Client) error {
 	for _, v := range data {
 		table.Append(v)
 	}
+	table.SetRowLine(true)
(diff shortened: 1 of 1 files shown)


Ankurk99 (approved):

probe/probe.go
```diff
@@ -625,7 +626,7 @@ func getPoliciesOnAnnotatedPods(c *k8s.Client) (map[string][]string, error) {
 			}
 		}
 	} else {
-		fmt.Printf("No Resource found in namespace\n")
+		fmt.Printf("No kubearmor policies found\n")
 	}
```
daemon1024, line 630: Let's not keep this. Like empty policy section in the below section anyway means that there no policies right?

Essietom: right!

PR #155, opened 2022-08-30
Update karmor recommend command with user labels

vishnusomank:
- Added `use-labels` flag to input labels for policy
- Updated the Options struct to include `Uselabels` field
- Updated policy logic with `ownerOnly` flag and user-defined labels

Addresses: https://github.com/kubearmor/kubearmor-client/issues/112#issuecomment-1216476912


recommend/policy.go
```diff
@@ -89,7 +90,15 @@ func (img *ImageInfo) createPolicy(ms MatchSpec) (types.KubeArmorPolicy, error)
 
 	// add container selector
 	repotag := strings.Split(img.RepoTags[0], ":")
-	policy.Spec.Selector.MatchLabels["kubearmor.io/container.name"] = repotag[0]
+	// If user defined labels are present, update the matchLabels with them or use default matchLabels
+	if len(options.Uselabels) > 0 {
+		for _, uselabel := range options.Uselabels {
+			userLabel := strings.Split(uselabel, "=")
```
nyrahul, line 96: Can we split either on the basis of "=" or ":' ?

I have seen folks getting confused as to what separator to use and the code can support either.

recommend/policy.go
```diff
@@ -37,6 +37,7 @@ func addPolicyRule(policy *types.KubeArmorPolicy, r Rules) {
 			dirRule := types.KnoxMatchDirectories{
 				Dir:        path,
 				Recursive:  true,
+				OwnerOnly:  true,
```
nyrahul, line 40: By default, this should be false.
```suggestion
				OwnerOnly:  false,
```

nyrahul (changes requested):

recommend/policy.go
vishnusomank: Yes, I've updated the code to include both "=" and ":"

recommend/policy.go
vishnusomank: @nyrahul  When using `OwnerOnly: false`  the entire `ownerOnly` part is getting omitted in the policy YAML. From my research, I could see that to include false boolean data in the output file,  we have to use pointer booleans [https://stackoverflow.com/questions/37756236/json-golang-boolean-omitempty]

vishnusomank: @nyrahul The current code will add the ownerOnly flag if it's enabled in Rules.json. I hope this will address the issue we had.
1. Output with ownerOnly: true in Rules.json
    ```
      cat out/alpine-latest/sys-bin-protect.yaml
      apiVersion: security.kubearmor.com/v1
      kind: KubeArmorPolicy
      metadata:
        name: sys-bin-protect
      spec:
        action: Audit
        file:
          matchDirectories:
          - dir: /bin/
            ownerOnly: true
            recursive: true
          - dir: /usr/bin/
            ownerOnly: true
            recursive: true
          - dir: /usr/sbin/
            ownerOnly: true
            recursive: true
        message: attempted access to system binaries
        selector:
          matchLabels:
            mysql: testdata
            test: labels
            testvalue: isgreat
        severity: 1
        tags:
        - CIS
        - MITRE
     ```
2. Output without `ownerOnly: true` in Rules.json
    ```
    cat out/alpine-latest/maint-tool-access.yaml
    apiVersion: security.kubearmor.com/v1
    kind: KubeArmorPolicy
    metadata:
      name: maint-tool-access
    spec:
      action: Audit
      file:
        matchDirectories:
        - dir: /sbin/
          recursive: true
      message: restricted maintenance tool access attempted
      selector:
        matchLabels:
          mysql: testdata
          test: labels
          testvalue: isgreat
      severity: 1
      tags:
      - PCI-DSS
      - MITRE
    ```

nyrahul (changes requested):
Can you please fix all the lint issues (existing as well as the new ones)? Thanks

vishnusomank:
Hi, @nyrahul I've updated the code to fix all the lint issues. Please review.

nyrahul (approved):
Many Thanks @vishnusomank for taking care of fixing the lint issues introduced by other committers. May be we should fail the CI if any new lint warning is introduced! WDYT? @daemon1024

PR #157, opened 2022-09-01
recommend: Converted rules spec file to YAML format

wazir-ahmed:


nyrahul (approved):
This is awesome. Exactly what I was looking for. 👍🏼

PR #159, opened 2022-09-01
recommend: Added processRule support

wazir-ahmed:

### DIFF
--- a/recommend/yaml/rules.yaml
+++ b/recommend/yaml/rules.yaml
@@ -126,3 +126,48 @@ policyRules:
     - PCI-DSS
     - MITRE
     action: Audit
+- name: shell-access
+  precondition: "/bin/sh"
+  description:
+    refs:
+    - name: MITRE-TTP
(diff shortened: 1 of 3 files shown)


recommend/yaml/rules.yaml
```diff
@@ -115,7 +115,7 @@ policyRules:
       never be used in prod env, or if used, should be used only in certain time frames.
       Examples include, dynamic package management tools, mii-tool, iptables etc
   rules:
-    pathRule:
+    fileRule:
```
nyrahul, line 118: Can you please add an example processRule in the rules yaml? Thanks

nyrahul (changes requested):

recommend/yaml/rules.yaml
wazir-ahmed: @nyrahul Done - eca7114

nyrahul (approved):

PR #165, opened 2022-09-05
use go 1.18 buildinfo for version output

### LINKED ISSUE #55: Use go 1.18 buildinfo for version output
go 1.18 introduced a new `debug/buildinfo` package which should eliminate the need of manually embedding version info using build flags.

Let's use it for our versioning information

Ref https://go.dev/doc/go1.18#debug/buildinfo

osBins:
Work in progress. To fix #55


daemon1024:
Thanks for the PR. Switching this to draft since it's still WIP. Please revert it back to normal PR when you think it's ready 😄

PR #166, opened 2022-09-05
probe: update kernel header check

daemon1024:
- We now check for either BTF Information or Kernel Headers.
- recreate the daemonset without relevant mounts  if error while mounting kernel headers probe
- refactored exec into pod logic


daemon1024:
@Essietom Can you check out and review the PR please? Thank You

Essietom (approved):

PR #171, opened 2022-09-08
Enable aggregation in kubearmor summary

seswarrajan:

### DIFF
--- a/summary/table.go
+++ b/summary/table.go
@@ -7,7 +7,6 @@ import (
 	"strings"
 
 	opb "github.com/accuknox/auto-policy-discovery/src/protobuf/v1/observability"
-	"github.com/rs/zerolog/log"
 
 	"github.com/olekukonko/tablewriter"
 )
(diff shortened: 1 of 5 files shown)


summary/table.go
```diff
@@ -23,6 +24,11 @@ var (
 // DisplaySummaryOutput function
 func DisplaySummaryOutput(resp *opb.Response, revDNSLookup bool, requestType string) {
 
+	if len(resp.ProcessData) <= 0 && len(resp.FileData) <= 0 && len(resp.InNwData) <= 0 && len(resp.OutNwData) <= 0 {
+		log.Error().Msgf("no summary data available")
+		return
+	}
```
daemon1024, line 30: Ideally DisplaySummaryOutput should return the Error, then this error will be handled by Cobra itself!

daemon1024:
One concern, Rest LGTM 👌🏽

summary/table.go
seswarrajan: Hello @daemon1024 
Since this is running in a loop, It is not required to throw an error. It is possible to get some values as empty. This error is actually an info.

daemon1024 (approved):

PR #173, opened 2022-09-08
recommend: Support for k8s manifest

wazir-ahmed:
- Fetch deployments from k8s based on given namespace & labels
- Get the images used by the deployment
- Generate policy for each image based on rules.yaml

### DIFF
--- a/cmd/recommend.go
+++ b/cmd/recommend.go
@@ -4,8 +4,6 @@
 package cmd
 
 import (
-	"errors"
-
 	"github.com/kubearmor/kubearmor-client/recommend"
 	"github.com/spf13/cobra"
 )
@@ -18,10 +16,6 @@ var recommendCmd = &cobra.Command{
 	Short: "Recommend Policies",
 	Long:  `Recommend policies based on container image, k8s manifest or the actual runtime env`,
 	RunE: func(cmd *cobra.Command, args []string) error {
-		//Condition to check if at least one Container image name is passes as an argument
-		if len(recommendOptions.Images) < 1 {
-			return errors.New("at least one container image is required as an argument")
-		}
 		if err := recommend.Recommend(client, recommendOptions); err != nil {
 			return err
 		}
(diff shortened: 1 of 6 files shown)


recommend/recommend.go
```diff
@@ -83,23 +94,83 @@ func Recommend(c *k8s.Client, o Options) error {
...
+				selectorLabels = append(selectorLabels, fmt.Sprintf("%s=%s", k, v))
+			}
+
+			matched := true
+			for _, filterLabel := range o.Labels {
+				if !slices.Contains(selectorLabels, filterLabel) {
+					matched = false
+					break
+				}
+			}
```
nyrahul, line 120: There is no need to create another array here.
Since MatchLabels is a map indexed by label key, you can directly use the input key to check whether the label exists and compare it.
Also it would be good to have a function like `matchLabels()` used here so that the main logic does not look so complicated.

recommend/report_text.go
```diff
@@ -59,7 +60,7 @@ func (r TextReport) SectionEnd(img *ImageInfo) error {
 func (r TextReport) Record(ms MatchSpec, policyName string) error {
 	var rec []string
 
-	rec = append(rec, policyName)
+	rec = append(rec, wrapPolicyName(policyName, 35))
```
nyrahul, line 63: It is possible to set the column width with TableWriter. ([ref](https://github.com/olekukonko/tablewriter/blob/74c60be0ef6886d331ebc3c91b2ac0a7971715b6/table.go#L194))

nyrahul (changes requested):

recommend/report_text.go
wazir-ahmed: I fought with the settings available in `TableWriter` before writing this function. I found out `TableWriter` wraps the text to the next line only after end of a word (when encountering a `space` char). Since `policyName` is a single word, it doesn't gets wrapped and the columns gets misaligned in the CLI output. That's why I ended up writing this function.

cmd/recommend.go
```diff
@@ -18,10 +16,10 @@ var recommendCmd = &cobra.Command{
...
 	Long:  `Recommend policies based on container image, k8s manifest or the actual runtime env`,
 	RunE: func(cmd *cobra.Command, args []string) error {
-		//Condition to check if at least one Container image name is passes as an argument
-		if len(recommendOptions.Images) < 1 {
-			return errors.New("at least one container image is required as an argument")
-		}
+		// Condition to check if at least one Container image name is passes as an argument
+		// if len(recommendOptions.Images) < 1 {
+		// 	return errors.New("at least one container image is required as an argument")
+		// }
```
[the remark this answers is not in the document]
wazir-ahmed: Addressed.

recommend/policy.go
```diff
@@ -207,10 +199,10 @@ func (img *ImageInfo) getPolicyFromImageInfo() {
 			continue
 		}
 
-		poldir := fmt.Sprintf("%s/%s", options.OutDir, mkPathFromTag(img.RepoTags[0]))
+		poldir := img.getPolicyDir()
 		_ = os.Mkdir(poldir, 0750)
 
-		outfile := fmt.Sprintf("%s/%s.yaml", poldir, policy.Metadata["name"])
+		outfile := img.getPolicyFile(ms.Name)
```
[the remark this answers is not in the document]
wazir-ahmed: Done.

recommend/recommend.go
```diff
@@ -83,23 +94,83 @@ func Recommend(c *k8s.Client, o Options) error {
 		ReportInit(o.ReportFile)
 	}
 
-	o.Images = unique(o.Images)
+	if len(o.Images) == 0 {
+		// recommendation based on k8s manifest that
+		// matches the given namespace and labels
+		if o.Namespace == "" {
```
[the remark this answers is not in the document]
wazir-ahmed: Done.

recommend/recommend.go
wazir-ahmed: Done.

nyrahul (approved):
LGTM 👍🏼

PR #178, opened 2022-09-12
Display PodInfo in table format

seswarrajan:

### DIFF
--- a/summary/table.go
+++ b/summary/table.go
@@ -27,9 +27,7 @@ func DisplaySummaryOutput(resp *opb.Response, revDNSLookup bool, requestType str
 		return
 	}
 
-	podInfo := resp.PodName + "/" + resp.Namespace + "/" + resp.ClusterName + "/" + resp.Label + "/" + resp.ContainerName
-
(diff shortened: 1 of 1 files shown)


summary/table.go
```diff
@@ -136,3 +134,27 @@ func WriteTable(header []string, data [][]string) {
...
+	fmt.Printf("\n")
+
+	podinfo := [][]string{
+		{"Pod Name", podname},
+		{"Namespace Name", namespace},
+		{"Cluster Name", clustername},
+		{"Container Name", containername},
+		{"Labels", labels},
+	}
+	table := tablewriter.NewWriter(os.Stdout)
```
daemon1024, line 149: We have tablewriter function in probe and recommend as well. Can we check if the following code as refactored?

seswarrajan: Hello @daemon1024 
This file also contains a tableWriter function. But since I wanted to write this info with no border, I wrote a new function.

nyrahul (approved):
LGTM 👍🏼

PR #179, opened 2022-09-12
recommend runtime policy using karmor

vishnusomank:
- Added the ability to query the discovery engine to get summary details
- Added the ability to create dynamic policy based on k8s service account access data from the discovery engine
- Added support for the creation of a single policy with multiple rules
- Optimised the function to include details of generated policies on the report file

Addresses https://github.com/kubearmor/kubearmor-client/issues/112#issuecomment-1216476912

### DIFF
--- a/recommend/policy.go
+++ b/recommend/policy.go
@@ -9,9 +9,8 @@ import (
 	"path/filepath"
 	"strings"
 
-	"github.com/clarketm/json"
-
 	"github.com/accuknox/auto-policy-discovery/src/types"
+	"github.com/clarketm/json"
 	"github.com/fatih/color"
 	log "github.com/sirupsen/logrus"
 	"k8s.io/utils/strings/slices"
@@ -22,17 +21,20 @@ func (r *SysRule) convertToKnoxRule() types.KnoxSys {
 	knoxRule := types.KnoxSys{}
 	fromSourceArr := []types.KnoxFromSource{}
 
-	if r.FromSource != "" {
-		if strings.HasSuffix(r.FromSource, "/") {
-			fromSourceArr = append(fromSourceArr, types.KnoxFromSource{
-				Dir: r.FromSource,
-			})
-		} else {
-			fromSourceArr = append(fromSourceArr, types.KnoxFromSource{
-				Path: r.FromSource,
-			})
+	for _, eachSource := range r.FromSource {
+		if eachSource != "" {
+			if strings.HasSuffix(eachSource, "/") {
+				fromSourceArr = append(fromSourceArr, types.KnoxFromSource{
+					Dir: eachSource,
+				})
+			} else {
+				fromSourceArr = append(fromSourceArr, types.KnoxFromSource{
+					Path: eachSource,
+				})
+			}
 		}
 	}
+
 	for _, path := range r.Path {
 		if strings.HasSuffix(path, "/") {
 
@@ -161,6 +163,40 @@ func matchTags(ms MatchSpec) bool {
 	return false
 }
 
+func (img *ImageInfo) writePolicyFile(ms MatchSpec) {
+	policy, err := img.createPolicy(ms)
+	if err != nil {
+		log.WithError(err).WithFields(log.Fields{
+			"image": img, "spec": ms,
+		}).Error("create policy failed, skipping")
+
+	}
+
+	outFile := img.getPolicyFile(ms.Name)
+	_ = os.MkdirAll(filepath.Dir(outFile), 0750)
+
+	f, err := os.Create(filepath.Clean(outFile))
+	if err != nil {
+		log.WithError(err).Error(fmt.Sprintf("create file %s failed", outFile))
+
+	}
+
+	arr, _ := json.Marshal(policy)
+	yamlArr, _ := yaml.JSONToYAML(arr)
+	if _, err := f.WriteString(string(yamlArr)); err != nil {
+		log.WithError(err).Error("WriteString failed")
+	}
+	if err := f.Sync(); err != nil {
+		log.WithError(err).Error("file sync failed")
+	}
+	if err := f.Close(); err != nil {
+		log.WithError(err).Error("file close failed")
+	}
+	_ = ReportRecord(ms, outFile)
+	color.Green("created policy %s ...", outFile)
+
+}
+
 func (img *ImageInfo) getPolicyFromImageInfo() {
 	if img.OS != "linux" {
 		color.Red("non-linux platforms are not supported, yet.")
@@ -171,7 +207,16 @@ func (img *ImageInfo) getPolicyFromImageInfo() {
 		log.WithError(err).Error("report start failed")
 		return
 	}
-	ms, err := getNextRule(&idx)
+	var ms MatchSpec
+	var err error
+
+	err = createRuntimePolicy(img)
+	if err != nil {
+		log.WithError(err).Error("Failed to create runtime policy")
+		return
+	}
+
+	ms, err = getNextRule(&idx)
 	for ; err == nil; ms, err = getNextRule(&idx) {
 		// matches preconditions
 
@@ -182,37 +227,8 @@ func (img *ImageInfo) getPolicyFromImageInfo() {
 		if !img.checkPreconditions(ms) {
 			continue
 		}
-
-		policy, err := img.createPolicy(ms)
-		if err != nil {
-			log.WithError(err).WithFields(log.Fields{
-				"image": img, "spec": ms,
-			}).Error("create policy failed, skipping")
-			continue
-		}
-
-		outFile := img.getPolicyFile(ms.Name)
-		_ = os.MkdirAll(filepath.Dir(outFile), 0750)
-
-		f, err := os.Create(filepath.Clean(outFile))
-		if err != nil {
-			log.WithError(err).Error(fmt.Sprintf("create file %s failed", outFile))
-			continue
-		}
-
-		arr, _ := json.Marshal(policy)
-		yamlArr, _ := yaml.JSONToYAML(arr)
-		if _, err := f.WriteString(string(yamlArr)); err != nil {
-			log.WithError(err).Error("WriteString failed")
-		}
-		if err := f.Sync(); err != nil {
-			log.WithError(err).Error("file sync failed")
-		}
-		if err := f.Close(); err != nil {
-			log.WithError(err).Error("file close failed")
-		}
-		_ = ReportRecord(ms, outFile)
-		color.Green("created policy %s ...", outFile)
+		img.writePolicyFile(ms)
 	}
+
 	_ = ReportSectEnd(img)
 }
--- a/recommend/recommend.go
+++ b/recommend/recommend.go
@@ -106,7 +106,8 @@ func Recommend(c *k8s.Client, o Options) error {
 			return err
 		}
 		for _, dp := range dps.Items {
-			if matchLabels(labelMap, dp.Spec.Selector.MatchLabels) {
+
+			if matchLabels(labelMap, dp.Spec.Template.Labels) {
 				images := []string{}
 				for _, container := range dp.Spec.Template.Spec.Containers {
 					images = append(images, container.Image)
@@ -115,7 +116,7 @@ func Recommend(c *k8s.Client, o Options) error {
 				deployments = append(deployments, Deployment{
 					Name:      dp.Name,
 					Namespace: dp.Namespace,
-					Labels:    dp.Spec.Selector.MatchLabels,
+					Labels:    dp.Spec.Template.Labels,
 					Images:    images,
 				})
 			}
@@ -144,6 +145,7 @@ func Recommend(c *k8s.Client, o Options) error {
 }
 
 func handleDeployment(dp Deployment) error {
+
 	var err error
 	for _, img := range dp.Images {
 		tempDir, err = os.MkdirTemp("", "karmor")
--- a/recommend/runtimePolicy.go
+++ b/recommend/runtimePolicy.go
@@ -1,0 +1,100 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2022 Authors of KubeArmor
+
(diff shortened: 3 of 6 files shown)


recommend/runtimePolicy.go
```diff
@@ -0,0 +1,150 @@
...
+
+			}
+		}
+
+	}
+
+	if len(sumResp.FileData) > 0 {
+
+		for _, fileData := range sumResp.FileData {
+			if strings.Contains(fileData.ProcName, "/run/secrets/kubernetes.io/serviceaccount") {
```
wazir-ahmed, line 136: Define the service account file path as a `const` and use it here.

recommend/runtimePolicy.go
```diff
@@ -0,0 +1,150 @@
...
+		ms.Rules.FileRule = &filePaths
+
+	} else {
+		ms.Rules.ProcessRule = &processPaths
+
+	}
+
+	ms.OnEvent.Severity = 1
+	ms.Description.Tldr = "Kubernetes serviceaccount folder access should be limited "
+	ms.Precondition = "/run/secrets/kubernetes.io/serviceaccount"
```
wazir-ahmed, line 106: Define the service account file path as a `const` and use it here.

recommend/runtimePolicy.go
```diff
@@ -0,0 +1,150 @@
...
+	"strings"
+
+	opb "github.com/accuknox/auto-policy-discovery/src/protobuf/v1/observability"
+
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials/insecure"
+)
+
+var ms MatchSpec
+var filePaths, processPaths SysRule
```
wazir-ahmed, line 20: These global variable are unnecessary. As much as possible, try to create objects in the function scope and return them to caller.

recommend/runtimePolicy.go
```diff
@@ -0,0 +1,150 @@
...
+
+	ms, err := summarySearch(label, o.Namespace)
+
+	img.writePolicyFile(ms)
+
+	return err
+
+}
+
+func summarySearch(labels, namespace string) (MatchSpec, error) {
```
wazir-ahmed, line 41: This function should not rely on global `ms` variable. It should create a `MatchSpec` object locally and return it.

recommend/runtimePolicy.go
```diff
@@ -0,0 +1,150 @@
...
+		}
+		_ = checkProcessFileData(sumResp)
+
+	}
+
+	return ms, nil
+
+}
+
+func createRuntimePolicy(policyType string) MatchSpec {
```
wazir-ahmed, line 94: This function should not rely on global `filePaths` and `processPaths` variables, it should receive them arguments. It should create a `MatchSpec` based on the arguments and return it.

recommend/runtimePolicy.go
```diff
@@ -0,0 +1,150 @@
...
+	ms.Name = "audit-serviceaccount-runtime"
+	ms.OnEvent.Tags = []string{"KUBERNETES", "SERVICE ACCOUNT", "RUNTIME POLICY"}
+	ms.OnEvent.Message = "serviceaccount access detected"
+	ms.OnEvent.Action = "Audit"
+
+	return ms
+
+}
+
+func checkProcessFileData(sumResp *opb.Response) error {
```
wazir-ahmed, line 116: This function should not rely on global `ms` variable. It should create a `MatchSpec` locally and return it.

recommend/runtimePolicy.go
```diff
@@ -0,0 +1,150 @@
...
+	processPaths = SysRule{}
+	filePaths = SysRule{}
+	if len(sumResp.ProcessData) > 0 {
+
+		for _, procData := range sumResp.ProcessData {
```
wazir-ahmed, line 121: Why are we checking the process data? Checking the file access data wouldn't be sufficient?

wazir-ahmed (changes requested):

recommend/policy.go
```diff
@@ -161,7 +161,41 @@ func matchTags(ms MatchSpec) bool {
...
+	_ = os.MkdirAll(filepath.Dir(outFile), 0750)
+
+	f, err := os.Create(filepath.Clean(outFile))
+	if err != nil {
+		log.WithError(err).Error(fmt.Sprintf("create file %s failed", outFile))
+
+	}
+
+	arr, _ := json.Marshal(policy)
+	yamlArr, _ := yaml.JSONToYAML(arr)
```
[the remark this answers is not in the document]
vishnusomank: I've changed it into a single step `yaml.Marshal(policy)` but the resulting policies had empty filerules, processrules or networkrules. So I've changed to back

recommend/recommend.go
```diff
@@ -115,7 +116,7 @@ func Recommend(c *k8s.Client, o Options) error {
 				deployments = append(deployments, Deployment{
 					Name:      dp.Name,
 					Namespace: dp.Namespace,
-					Labels:    dp.Spec.Selector.MatchLabels,
+					Labels:    dp.Spec.Template.Labels,
```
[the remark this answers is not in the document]
vishnusomank: Yes, `selector.matchlabels` only gives us the labels for the deployment. To get the labels that are assigned for pods we need to get the `template.labels` value.
My ubuntu deployment had only one label `app: ubuntu` but the pod has 2 `app: ubuntu test: ubuntu` and the resulting policy also had only one label `app:ubuntu` that's why I changed it into `templates.labels`

recommend/runtimePolicy.go
```diff
@@ -0,0 +1,150 @@
...
+
+var ms MatchSpec
+var filePaths, processPaths SysRule
+
+// CreateRuntimePolicy function generates runtime policy for service account
+func CreateRuntimePolicy(o *Options, img *ImageInfo) error {
+	ms = MatchSpec{}
+	var label string
+
+	for _, labels := range o.Labels {
```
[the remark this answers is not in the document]
vishnusomank: I used o.Labels over img.Labels because I was trying to generate something like `app=ubuntu,test=ubuntu` for getting the summary details for a specific pod

recommend/recommend.go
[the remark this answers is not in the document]
wazir-ahmed: :+1:

recommend/runtimePolicy.go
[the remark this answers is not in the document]
wazir-ahmed: 1) You should convert `img.Labels` to the format accepted `client.GetPodNames()`. Not `o.Labels`. `o.Labels` contains the user given labels in the CLI to filter the deployment. `img.Labels` contains the exact labels of the pod.

2) Since we're passing the `img` object to `summarySearch()`, we can move this label format conversion inside `summarySearch()`

recommend/policy.go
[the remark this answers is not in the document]
wazir-ahmed: Okay. We can investigate this later.

recommend/runtimePolicy.go
```diff
@@ -0,0 +1,126 @@
...
+	ms.Name = "audit-serviceaccount-runtime"
+	ms.OnEvent.Tags = []string{"KUBERNETES", "SERVICE ACCOUNT", "RUNTIME POLICY"}
+	ms.OnEvent.Message = "serviceaccount access detected"
+	ms.OnEvent.Action = "Audit"
+
+	return ms
+
+}
+
+func checkProcessFileData(sumResp *opb.Response, ms MatchSpec) (MatchSpec, error) {
```
wazir-ahmed, line 107: Why this function receives an empty `ms` just to carry it forward to another function?
1. Let's remove `ms` argument from both `createProcessFIleData()` and `createRuntimePolicy()`. 
2. Please make `createRuntimePolicy()` create new `ms` locally with all filled value and return it.

wazir-ahmed (changes requested):
Please check the unresolved conversations.

recommend/runtimePolicy.go
```diff
@@ -0,0 +1,94 @@
...
+	opb "github.com/accuknox/auto-policy-discovery/src/protobuf/v1/observability"
+
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials/insecure"
+)
+
+const serviceAccData = "/run/secrets/kubernetes.io/serviceaccount"
+
+// CreateRuntimePolicy function generates runtime policy for service account
+func CreateRuntimePolicy(o *Options, img *ImageInfo) error {
```
wazir-ahmed, line 22: 1. The first argument `(o *Options)` in this function is unused.
2. Since this function has become a one-liner [ `return summarySearch(img)` ], we could delete `summarySearch()` and handle its operation as part of `CreateRuntimePolicy()`.

wazir-ahmed (changes requested):

recommend/runtimePolicy.go
```diff
@@ -0,0 +1,86 @@
...
+	defer conn.Close()
+	client := opb.NewObservabilityClient(conn)
+	podResp, err := client.GetPodNames(context.Background(), &opb.Request{
+		Label:     labels,
+		NameSpace: img.Namespace,
+	})
+	if err != nil {
+		return err
+	}
+	for _, podName := range podResp.PodName {
```
wazir-ahmed, line 47: This `podName` here, is it the exact name of pod? If yes, then consider following scenario..

In the `observability.Summary()`, if we get the following data..
```
deployment_1 --> pod_1 | service_account_file --> accessed by --> binary_1
deployment_1 --> pod_2 | service_account_file --> accessed by --> binary_2
```
Since for each pod we're create a new `MatchSpec` and converting it to a policy, this would create two `MatchSpec` and since the `MatchSpec.Name` is the same, policy file name will be the same and last policy written to the disk will overwrite the first policy.

@vishnusomank Let me if this could happen.

vishnusomank: Yes, the `podName` is the exact name of the pod. And the scenario mentioned might happen since we are not differentiating them based on pod name. Ideally, the scenario is highly improbable, right? If I have 2 pods (ie rs =2) then their behavior also should be the same.
Should we add `runtime-PODNAME` as part of the policy name to avoid this scenario? @wazir-ahmed

wazir-ahmed: @vishnusomank No, we should not create multiple policy file. There should only be one policy file per deployment for service account access. What we should do is.. we should combine the `fromSource` information from all the pods (of a particular deployment) and create a policy like below.
```yaml
file:
  matchPaths:
  - path: "/run/secrets/kubernetes.io/serviceaccount"
    fromSource:
    - path: <binary_1>
    - path: <binary_2>
    - path: <binary_n>
```
@nyrahul This is the expected output, right?

wazir-ahmed (changes requested):

recommend/runtimePolicy.go
```diff
@@ -0,0 +1,89 @@
...
+	// create a client
+	conn, err := grpc.Dial(gRPC, grpc.WithTransportCredentials(insecure.NewCredentials()))
+	if err != nil {
+		return errors.New("could not connect to the server. Possible troubleshooting:\n- Check if discovery engine is running\n- Create a portforward to discovery engine service using\n\t\033[1mkubectl port-forward -n explorer service/knoxautopolicy --address 0.0.0.0 --address :: 9089:9089\033[0m\n[0m")
+	}
+	defer conn.Close()
+	client := opb.NewObservabilityClient(conn)
+	sumResp, err := client.Summary(context.Background(), &opb.Request{
+		Label:     labels,
+		NameSpace: img.Namespace,
```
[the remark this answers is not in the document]
wazir-ahmed: In [`kubearmor summary`](https://github.com/kubearmor/kubearmor-client/blob/main/summary/summary.go#L77), first it queries for the list of pods with a given label and namespace and then get the summary based on pod-names. Two way process. Like how you have previous written the code. 

You need to have an array of `sumResp` and call `checkProcessFileData()` with that.

recommend/runtimePolicy.go
```diff
@@ -0,0 +1,89 @@
...
+		for _, procPath := range uniqueProcName {
+			if strings.Contains(procPath, serviceAccData) {
+				filePaths.Path = append(filePaths.Path, procPath)
+				for _, fileData := range sumResp.FileData {
+					if procPath == fileData.ProcName {
+						filePaths.FromSource = append(filePaths.FromSource, fileData.ParentProcName)
+					}
+				}
+			}
+		}
```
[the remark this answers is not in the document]
vishnusomank: @wazir-ahmed  If I change the code then some paths are getting missed out.  if there is access to `/run/secrets/kubernetes.io/serviceaccount/namespace` it'll be missing from the policy since `/run/secrets/kubernetes.io/serviceaccount/namespace ≠ serviceAccData`

recommend/runtimePolicy.go
```diff
@@ -0,0 +1,103 @@
...
+					filePaths.Path = append(filePaths.Path, procPath)
+					for _, fileData := range eachResp.FileData {
+						if procPath == fileData.ProcName {
+							filePaths.FromSource = append(filePaths.FromSource, fileData.ParentProcName)
+						}
+					}
+				}
+			}
+		}
+	}
```
wazir-ahmed, line 98:
```suggestion
	 	for _, fileData := range eachResp.FileData {
			if strings.HasPrefix(fileData.ProcName, saPath[0]) || strings.HasPrefix(fileData.ProcName, saPath[1])  {
				filePaths.FromSource = append(filePaths.FromSource, fileData.ParentProcName)
			}
		}
	}
	filePaths.Path = saPath
	filePaths.Recursive = true

```

wazir-ahmed (changes requested):

wazir-ahmed:
@vishnusomank PR looks good. Please address the following.
1. The `go-fmt` CI test is failing. Fix that.
2. Squash all the commits in the PR into a single commit.

vishnusomank:
Thanks for reviewing @wazir-ahmed. I've made changes according to the comments.

wazir-ahmed (approved):

Issue #176, opened 2022-09-12
handling policy-templates 


nyrahul:
Prepare rules.yaml based on [policy-templates repo](https://github.com/kubearmor/policy-templates/).

- [ ] prepare sample metadata.yaml
- [ ] prepare release of policy-templates
- [ ] `karmor recommend --update`
- [ ] `karmor recommend` check whether latest policy-templates are available.

vishnusomank:
Hi, @nyrahul @wazir-ahmed I've raised a draft PR for the same can you take a look at it and suggest if any changes or improvements are needed?
PR : [https://github.com/kubearmor/kubearmor-client/pull/180](https://github.com/kubearmor/kubearmor-client/pull/180)

PR #180, opened 2022-09-13
karmor recommend: handling policy-templates

vishnusomank:
- Created sample metadata.yaml in [repo](https://github.com/vishnusomank/policy-templates/blob/main/MySQL/system/metadata.yaml)
- Created release of [policy-templates](https://github.com/vishnusomank/policy-templates/releases)
- Updated code with `karmor recommend --update`. This will download the latest policy-template release
- Updated code with check for whether latest policy-templates are available when using `karmor recommend`

Addresses https://github.com/kubearmor/kubearmor-client/issues/176

### DIFF
--- a/recommend/policyTeamplates.go
+++ b/recommend/policyTeamplates.go
@@ -1,0 +1,241 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2022 Authors of KubeArmor
+
+package recommend
+
+import (
+	"archive/zip"
+	"context"
+	"errors"
+	"fmt"
+	"os"
+	"path"
+	"path/filepath"
+	"runtime"
+	"strings"
+
+	"github.com/cavaliergopher/grab/v3"
+	"github.com/google/go-github/github"
+	log "github.com/sirupsen/logrus"
+)
+
+const (
+	org  = "vishnusomank"
+	repo = "policy-templates"
+	url  = "https://github.com/vishnusomank/policy-templates/archive/refs/tags/"
+)
+
+// userHome function returns users home directory
+func userHome() string {
+
+	if runtime.GOOS == "windows" {
+		home := os.Getenv("HOMEDRIVE") + os.Getenv("HOMEPATH")
+		if home == "" {
+			home = os.Getenv("USERPROFILE")
+		}
+		return home
+	}
+	return os.Getenv("HOME")
+
+}
+
+// IsTemplates function returns error if policy-template folder not found
+func isTemplates() error {
+	path := fmt.Sprintf("%s/.cache/karmor/", userHome())
+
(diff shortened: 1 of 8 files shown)


vishnusomank:
Hi, @nyrahul @wazir-ahmed I've raised a draft PR for `recommend: handling policy-templates` can you take a look at it and suggest if any changes or improvements are needed?
PR : [https://github.com/kubearmor/kubearmor-client/pull/180](https://github.com/kubearmor/kubearmor-client/pull/180)

recommend/policyTeamplates.go
```diff
@@ -0,0 +1,241 @@
...
+		name, err := sanitizeArchivePath(dest, file.Name)
+		if err != nil {
+			return err
+		}
+		_ = os.MkdirAll(path.Dir(name), 0750)
+		create, err := os.Create(filepath.Clean(name))
+		if err != nil {
+			return err
+		}
+
```
nyrahul, line 159: use defer create.Close() here .. otherwise it leads to leak below in the case of error situation

recommend/policyTeamplates.go
```diff
@@ -0,0 +1,241 @@
...
+	read, err := zip.OpenReader(source)
+	if err != nil {
+		return err
+	}
+	defer read.Close()
+	for _, file := range read.File {
+		if file.Mode().IsDir() {
+			continue
+		}
+		open, err := file.Open()
```
nyrahul, line 146: open.Close() needs to be handled.

recommend/policyTeamplates.go
```diff
@@ -0,0 +1,241 @@
...
+}
+
+func updatePolicyRules(filePath string) error {
+
+	var files []string
+
+	err := filepath.Walk(filePath, func(path string, info os.FileInfo, err error) error {
+
+		if err != nil {
+
```
nyrahul, line 198:
```suggestion
```

recommend/policyTeamplates.go
```diff
@@ -0,0 +1,241 @@
...
+			log.WithError(err).Error("file sync failed")
+		}
+
+	}
+	return nil
+
+}
+
+func updatePolicyRules(filePath string) error {
+
```
nyrahul, line 192:
```suggestion
```

nyrahul (changes requested):
I think the important question for me from the PR is:

We are trying to create ~/.cache/karmor/rules.yaml in the user's home directory. Is that our intention? I dont think that is what we discussed!

My understanding was that `karmor update` would be used only by the developer to update the rules.yaml in the kubearmor-client repo. So after the update is done the rules.yaml would be updated and pushed to the repo as a PR.

CC: @wazir-ahmed WDYT?

PR #182, opened 2022-09-17
karmor recommend: handling policy-templates

vishnusomank:
- Updated `policy recommend` with `--update` flag
- Added functions to check for new policy-template releases
- Added functions to generate rules.yaml from metadata.yaml in policy-templates
- Updated policy creation logic will select embeded rules.yaml if it cannot generate rules.yaml from policy-templates
- Updated rules structure to include policy-template policies
- Updated runtime policy generation logic
  - added ability to continue generate policies from rules.yaml even if runtime policy generation fails

Fixes: https://github.com/kubearmor/kubearmor-client/issues/176

Signed-off-by: vishnusomank vishnu@accuknox.com

### DIFF
--- a/cmd/recommend.go
+++ b/cmd/recommend.go
@@ -4,7 +4,10 @@
 package cmd
 
 import (
+	"fmt"
+
 	"github.com/kubearmor/kubearmor-client/recommend"
+	log "github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 )
 
@@ -22,9 +25,26 @@ var recommendCmd = &cobra.Command{
 		return nil
 	},
 }
+var updateCmd = &cobra.Command{
+	Use:   "update",
+	Short: "Updates policy-template cache",
+	Long:  "Updates the local cache of policy-templates ($HOME/.cache/karmor)",
+	RunE: func(cmd *cobra.Command, args []string) error {
+
+		if d, err := recommend.DownloadAndUnzipRelease(); err != nil {
+			fmt.Println(d)
+			return err
+		}
+		log.WithFields(log.Fields{
+			"Current Version": recommend.CurrentVersion,
+		}).Info("policy-templates updated")
+		return nil
+	},
+}
 
 func init() {
 	rootCmd.AddCommand(recommendCmd)
+	recommendCmd.AddCommand(updateCmd)
 
 	recommendCmd.Flags().StringSliceVarP(&recommendOptions.Images, "image", "i", []string{}, "Container image list (comma separated)")
 	recommendCmd.Flags().StringSliceVarP(&recommendOptions.Labels, "labels", "l", []string{}, "User defined labels for policy (comma separated)")
--- a/recommend/policy.go
+++ b/recommend/policy.go
@@ -9,93 +9,24 @@ import (
 	"path/filepath"
 	"strings"
 
-	"github.com/accuknox/auto-policy-discovery/src/types"
 	"github.com/clarketm/json"
 	"github.com/fatih/color"
+	pol "github.com/kubearmor/KubeArmor/pkg/KubeArmorPolicy/api/security.kubearmor.com/v1"
 	log "github.com/sirupsen/logrus"
 	"k8s.io/utils/strings/slices"
 	"sigs.k8s.io/yaml"
 )
 
-func (r *SysRule) convertToKnoxRule() types.KnoxSys {
-	knoxRule := types.KnoxSys{}
-	fromSourceArr := []types.KnoxFromSource{}
+func addPolicyRule(policy *pol.KubeArmorPolicy, r pol.KubeArmorPolicySpec) {
 
-	for _, eachSource := range r.FromSource {
-		if eachSource != "" {
-			if strings.HasSuffix(eachSource, "/") {
-				fromSourceArr = append(fromSourceArr, types.KnoxFromSource{
-					Dir: eachSource,
-				})
-			} else {
-				fromSourceArr = append(fromSourceArr, types.KnoxFromSource{
-					Path: eachSource,
-				})
-			}
-		}
+	if len(r.File.MatchDirectories) != 0 || len(r.File.MatchPaths) != 0 {
+		policy.Spec.File = r.File
 	}
-
-	for _, path := range r.Path {
-		if strings.HasSuffix(path, "/") {
-
-			dirRule := types.KnoxMatchDirectories{
-				Dir:        path,
-				Recursive:  r.Recursive,
-				FromSource: fromSourceArr,
-				OwnerOnly:  r.OwnerOnly,
-			}
-			knoxRule.MatchDirectories = append(knoxRule.MatchDirectories, dirRule)
-		} else {
-			pathRule := types.KnoxMatchPaths{
-				Path:       path,
-				FromSource: fromSourceArr,
-				OwnerOnly:  r.OwnerOnly,
-			}
-			knoxRule.MatchPaths = append(knoxRule.MatchPaths, pathRule)
-		}
+	if len(r.Process.MatchDirectories) != 0 || len(r.Process.MatchPaths) != 0 {
+		policy.Spec.Process = r.Process
 	}
-
-	return knoxRule
-}
-
-// networkToKnoxRule function to include KubeArmor network rules
-func (r *NetRule) networkToKnoxRule() types.NetworkRule {
-	knoxNetRule := types.NetworkRule{}
-	fromSourceArr := []types.KnoxFromSource{}
-
-	if r.FromSource != "" {
-		if strings.HasSuffix(r.FromSource, "/") {
-			fromSourceArr = append(fromSourceArr, types.KnoxFromSource{
-				Dir: r.FromSource,
-			})
-		} else {
-			fromSourceArr = append(fromSourceArr, types.KnoxFromSource{
-				Path: r.FromSource,
-			})
-		}
-	}
-	for _, protocol := range r.Protocol {
-
-		protoRule := types.KnoxMatchProtocols{
-			Protocol:   protocol,
-			FromSource: fromSourceArr,
-		}
-		knoxNetRule.MatchProtocols = append(knoxNetRule.MatchProtocols, protoRule)
-
-	}
-
-	return knoxNetRule
-}
-
-func addPolicyRule(policy *types.KubeArmorPolicy, r Rules) {
-	if r.FileRule != nil {
-		policy.Spec.File = r.FileRule.convertToKnoxRule()
-	}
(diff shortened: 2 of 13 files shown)


nyrahul (approved):

recommend/policyTemplates.go
```diff
@@ -0,0 +1,247 @@
...
+				skipCount := len(ms.Yaml) + 6
+				tempYaml = tempYaml[strings.Index(string(tempYaml), "yaml:")+skipCount:]
+				dataVal := strings.TrimSpace(string(tempYaml))
+				yamlFile = yamlFile[:strings.Index(string(yamlFile), "yaml:")-1]
+				for _, eachLine := range strings.Split(string(newYaml[strings.Index(string(newYaml), "spec:"):]), "\n") {
+					yamlFile = append(yamlFile, []byte(fmt.Sprintf(" %s\n", eachLine))...)
+				}
+				yamlFile = []byte(strings.TrimSpace(string(yamlFile)))
+				yamlFile = append(yamlFile, []byte(fmt.Sprintf("\n%s", dataVal))...)
+			}
```
wazir-ahmed, line 234: We shouldn't get our hands dirty by parsing the YAML file manually. Each `ms.Yaml` should be a valid KubeArmorPolicy file, right? So, use `KubeArmorPolicy` struct from [the package](https://pkg.go.dev/github.com/kubearmor/KubeArmor/pkg/KubeArmorPolicy/api/security.kubearmor.com/v1), do `yaml.Unmarshall` and extract the `spec` from the policy.

wazir-ahmed (changes requested):
@vishnusomank This PR is a big upgrade to karmor-recommend. Thanks! Could you please address the comments.

recommend/policyTemplates.go
```diff
@@ -0,0 +1,247 @@
...
+			home = os.Getenv("USERPROFILE")
+		}
+		return home
+	}
+	return os.Getenv("HOME")
+}
+
+// isTemplates function returns error if policy-template folder not found
+func isTemplates() error {
+	path := fmt.Sprintf("%s/.cache/karmor/", userHome())
```
wazir-ahmed, line 45: Keep the `.cache/karmor/` as a constant. Its better to write a function like `getCachePath()` and use it all the places.

recommend/policyTemplates.go
```diff
@@ -0,0 +1,247 @@
...
+	if stat, err := os.Stat(path); err == nil && stat.IsDir() {
+		err := isLatest()
+		if err != nil {
+			if selfupdate.ConfirmUserAction("Outdated policy-templates detected. Do you want to update it?") {
```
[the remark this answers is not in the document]
nyrahul: I agree that getting user input should not be done. Just showing a warning and continuing should be the way to go.

recommend/policyTemplates.go
vishnusomank: Since all the policies in policy-template start with a comment, it's showing JSON unmarshal error `error="invalid character '#' looking for beginning of value"` @wazir-ahmed

recommend/policy.go
```diff
@@ -17,85 +17,16 @@ import (
...
-		}
-	}
-	for _, protocol := range r.Protocol {
-
-		protoRule := types.KnoxMatchProtocols{
-			Protocol:   protocol,
-			FromSource: fromSourceArr,
-		}
-		knoxNetRule.MatchProtocols = append(knoxNetRule.MatchProtocols, protoRule)
+func addPolicyRule(policy *types.KnoxSystemPolicy, r types.KnoxSystemSpec) {
```
wazir-ahmed, line 20: Please don't use pkg from auto-policy-discovery. Use the pkg from KuberArmor repo - https://pkg.go.dev/github.com/kubearmor/KubeArmor/pkg/KubeArmorPolicy/api/security.kubearmor.com/v1

recommend/policyRules.go
```diff
@@ -11,17 +11,18 @@ import (
 	"github.com/clarketm/json"
 	"sigs.k8s.io/yaml"
 
+	"github.com/accuknox/auto-policy-discovery/src/types"
```
wazir-ahmed, line 14: Use https://pkg.go.dev/github.com/kubearmor/KubeArmor/pkg/KubeArmorPolicy/api/security.kubearmor.com/v1

recommend/policyTemplates.go
```diff
@@ -0,0 +1,228 @@
...
+// isTemplates function returns error if policy-template folder not found
+func isTemplates() error {
+	path := fmt.Sprintf("%s/%s", userHome(), getCachePath())
+	if stat, err := os.Stat(path); err == nil && stat.IsDir() {
+		if !isLatest() {
+			log.Warn("\033[1;33mA new version of policy-templates is available. Use `karmor recommend --update` to get recommendations based on the latest policy-templates.\033[0m")
+		}
+	} else {
+		log.Warn("\033[1;33mNo version of policy-templates available locally. Use `karmor recommend --update` to get recommendations based on the latest policy-templates.\033[0m")
+	}
```
wazir-ahmed, line 58: For both the cases, i.e if the local cache doesn't exists OR cache is outdated, throw the same message..
```
A new version of policy-templates is available. Use `karmor recommend update` to get recommendations based on the latest policy-templates.
```

recommend/runtimePolicy.go
```diff
@@ -61,40 +63,65 @@ func createRuntimePolicy(img *ImageInfo) error {
...
+					})
+				}
 			}
 		}
 	}
-	filePaths.Path = saPath
-	filePaths.Recursive = true
+	if len(fromSourceArr) < 1 {
+		log.Info("No serviceaccount access detected. Skipping runtime policy creation on ", distro)
+		return &MatchSpec{}, nil
```
wazir-ahmed, line 106: Creating an empty `MatchSpec` object and returning the pointer of that object doesn't make any sense. You should use `nil` here and handle the nil-case in the parent function.

recommend/yaml/rules.yaml
```diff
@@ -1,6 +1,6 @@
-policyRules:
```
wazir-ahmed, line 1: Add this line. The `rules.yaml` should start with..
```yaml
version: <semver>
policyRules:
- <MatchSpec-1>
- <MatchSpec-2>
```

recommend/policyTemplates.go
```diff
@@ -0,0 +1,228 @@
...
+}
+func currentRelease() string {
+	path := fmt.Sprintf("%s/%s", userHome(), getCachePath())
+	files, err := os.ReadDir(path)
+	if err != nil {
+		return "nil"
+	}
+	for _, file := range files {
+		if file.IsDir() {
+			currentVersion = fmt.Sprintf("v%s", strings.Split(file.Name(), "policy-templates-")[1])
```
wazir-ahmed, line 78: We should keep the store/retrieve the current version from `rules.yaml`.
```yaml
version: 0.0.1
policySpec:
- <MatchSpec-1>
- <MatchSpec-2>
```
In future, the generation of `rules.yaml` that ships with `karmor` binary by default, will be automated. The CI will automatically take the latest policy-template release, generate `rules.yaml`, set the `version` field to latest release version of policy-template and embedded that data into `karmor`. 

**(To-Do)**
If we want to incorporate this idea, then when we generate `rules.yaml` while `karmor recommend update`, we should add the version as part of `rules.yaml`. And when we call `updateRulesYAML` we should the parse the `version` and store it somewhere.

cmd/recommend.go
```diff
@@ -29,6 +29,7 @@ func init() {
 	recommendCmd.Flags().StringSliceVarP(&recommendOptions.Images, "image", "i", []string{}, "Container image list (comma separated)")
 	recommendCmd.Flags().StringSliceVarP(&recommendOptions.Labels, "labels", "l", []string{}, "User defined labels for policy (comma separated)")
 	recommendCmd.Flags().StringVarP(&recommendOptions.Namespace, "namespace", "n", "", "User defined namespace value for policies")
+	recommendCmd.Flags().BoolVar(&recommendOptions.Update, "update", false, "Updates the local cache of policy-templates ($HOME/.cache/karmor)")
```
wazir-ahmed, line 32: `--update` should not be a flag in `karmor recommend`. It should be a sub-command like `karmor recommend update`.

wazir-ahmed (changes requested):

recommend/policyTemplates.go
wazir-ahmed: If we keep `version` as part of `rules.yaml`, then even while using embedded `rules.yaml`, we will be able to compare the current version with the latest one and decide if this warning needs to be shown.

recommend/yaml/rules.yaml
```diff
@@ -1,6 +1,8 @@
+version: v0.1.3
```
wazir-ahmed, line 1: @vishnusomank Is this `0.1.3` represents something?

recommend/policy.go
```diff
@@ -214,8 +147,7 @@ func (img *ImageInfo) getPolicyFromImageInfo() {
 
 	err = createRuntimePolicy(img)
 	if err != nil {
-		log.WithError(err).Error("Failed to create runtime policy")
-		return
+		log.Info("No serviceaccount access detected. Skipping runtime policy creation on ", img.Distro)
```
wazir-ahmed, line 150:
```suggestion
		log.Infof("Failed to create runtime policy creation for %s/%s/%s. err=%s", img.Namespace, img.Deployment, img.Name, err)
```
The child functions of `createRuntimePolicy()` throw error for things like conn failure or RPC call failure. So we should print it here.

In case of skipping the policy creation, we just log that in `checkProcessFileData()`. So we're good in that case.

wazir-ahmed (changes requested):

recommend/yaml/rules.yaml
vishnusomank: it is v0.1.0 the current policy-template release

wazir-ahmed: As discussed privately, let's make this `0.0.1` since currently these are just sample policies.

wazir-ahmed (approved):
PR looks good to me.

recommend/policyTemplates.go
```diff
@@ -0,0 +1,236 @@
...
+	url   = "https://github.com/kubearmor/policy-templates/archive/refs/tags/"
+	cache = ".cache/karmor/"
+)
+
+// CurrentVersion stores the current version of policy-template
+var CurrentVersion string
+
+func getCachePath() string {
+	cache := fmt.Sprintf("%s/%s", userHome(), cache)
+	return cache
```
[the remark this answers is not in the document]
wazir-ahmed: Please add this as part of future PRs.

Issue #186, opened 2022-09-26
`karmor recommend` bugs/enhancements


daemon1024:
- [ ] Missing Policies in Report.
<details>
  <summary>We don't  have cert -access policy in report but the YAML is present in out directory</summary>

![image](https://user-images.githubusercontent.com/47106543/192247789-890ebca3-fbcc-4d30-a289-43154a800568.png)
</details>

- [ ] Use Policy Names instead of relative paths in Table writer. The directory path can be printed as part of other details prior to the Table thereby reducing clutter.
- [ ] Read `/etc/os-release` and provide additional preconditions options for distros and rules.
- [ ] Improve message for when there's no discovery engine. It should just simply mention that there's no runtime based recommendation instead of printing out the debug error.
- [ ] Potential improvement to Speed? It's understandable for the first time but it still takes a while for me on the subsequent calls for it to process and report things.

cc @nyrahul @vishnusomank

vishnusomank:
@daemon1024 
- Missing policies in the report: This issue happens because the report file (default report.txt) is getting replaced every time we run `karmor recommend`. So only the last updated policies and their details will be in the report file. Should we add time to the filename so it'll not be rewritten? cc: @nyrahul @wazir-ahmed 
- Preconditions are updated with the `metadata.yaml` file in the policy-templates repository. So is it necessary to read `/etc/os-release` and provide additional preconditions? cc: @nyrahul

PR #188, opened 2022-09-27
`karmor recommend` enhancement

vishnusomank:
- Updated table writer summary with policy-template version and output directory path
- Removed relative policy path from table writer to avoid clutter
- Performance improvement in policy recommendation

Fixes: https://github.com/kubearmor/kubearmor-client/issues/186


nyrahul (approved):
LGTM 👍🏼

PR #195, opened 2022-10-16
karmor recommend enhancements

vishnusomank:
- Updated message for serviceaccount deny policy with links to deny serviceaccount automount
- New flag `--config` to input the docker config to pull private repos (GCR,ECR,JFrog,Nexus,Docker private)
- Minor changes in report files (get severity, tags from rule level as well as policy level)

### DIFF
--- a/recommend/report_html.go
+++ b/recommend/report_html.go
@@ -107,6 +107,7 @@ func NewHTMLReport() HTMLReport {
 
 // Start of HTML report section
 func (r HTMLReport) Start(img *ImageInfo) error {
+
 	seci := SectionInfo{
 		HdrCols: []Col{
 			{Name: "Policy"},
@@ -118,7 +119,7 @@ func (r HTMLReport) Start(img *ImageInfo) error {
 		ImgInfo: []Info{
 			{Key: "Container", Val: img.RepoTags[0]},
 			{Key: "OS/Arch/Distro", Val: img.OS + "/" + img.Arch + "/" + img.Distro},
-			{Key: "Output Directory", Val: options.OutDir + "/" + img.Namespace + "-" + img.Deployment + "/"},
+			{Key: "Output Directory", Val: img.getPolicyDir()},
 			{Key: "policy-template version", Val: CurrentVersion},
 		},
 	}
@@ -136,7 +137,7 @@ type RecordInfo struct {
 }
 
 // Record addition of new HTML table row
-func (r HTMLReport) Record(ms MatchSpec, policyName string) error {
+func (r HTMLReport) Record(ms *MatchSpec, policyName string) error {
 	*r.RecordCnt = *r.RecordCnt + 1
 	policy, err := os.ReadFile(filepath.Clean(policyName))
 	if err != nil {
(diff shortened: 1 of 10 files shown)


recommend/report_html.go
```diff
@@ -136,20 +137,39 @@ type RecordInfo struct {
...
+		severity = fmt.Sprintf("%d", ms.Spec.Severity)
+	}
+	if ms.Spec.File.Action != "" {
+		action = string(ms.Spec.File.Action)
+	} else if ms.Spec.Process.Action != "" {
+		action = string(ms.Spec.Process.Action)
+	} else if ms.Spec.Network.Action != "" {
+		action = string(ms.Spec.Network.Action)
+	} else {
+		action = string(ms.Spec.Action)
```
nyrahul, line 164: We have a problem here.
Consider we have two rules, one file and another process, with Severity 1, 2 resp and Action as audit, block resp.

The severity will be decided to be 1 and action will be concluded to be audit.

I am not sure how to handle this.

We cannot show a single severity/action at the policy level if multiple rules have their individual severity/action.

recommend/report_text.go
```diff
@@ -63,13 +63,32 @@ func (r TextReport) SectionEnd(img *ImageInfo) error {
...
+		severity = fmt.Sprintf("%d", ms.Spec.Severity)
+	}
+	if ms.Spec.File.Action != "" {
+		action = string(ms.Spec.File.Action)
+	} else if ms.Spec.Process.Action != "" {
+		action = string(ms.Spec.Process.Action)
+	} else if ms.Spec.Network.Action != "" {
+		action = string(ms.Spec.Network.Action)
+	} else {
+		action = string(ms.Spec.Action)
```
nyrahul, line 85: This code is duplicated. It could have been a common API that could be called from text and html report code.

recommend/runtimePolicy.go
```diff
@@ -113,14 +114,45 @@ func checkProcessFileData(sumResp []*opb.Response, distro string) *MatchSpec {
 	ms.Spec = pol.KubeArmorPolicySpec{
 		Action:   "Allow",
 		Message:  "serviceaccount access detected",
-		Tags:     []string{"KUBERNETES", "SERVICE ACCOUNT", "RUNTIME POLICY"},
+		Tags:     []string{"KUBERNETES", "SERVICE ACCOUNT", "RUNTIME POLICY", "MITRE_T1552.007_container_api", "MITRE"},
```
nyrahul, line 117: This hardcoded policy rules in the code is not acceptable. Are we planning to handle it in different PR?

nyrahul (changes requested):

recommend/runtimePolicy.go
vishnusomank: @nyrahul we are having the runtime recommendation for serviceaccount as v1.
can we have something like the following
```
the user checkout the karmor summary and finds a list of destination files, and processes to secure. 
He uses karmor recommend --get-runtime "serviceaccount, .git, passwd" and in the backend, 
we split these keywords and check for them in logs using discovery-engine API and generate runtime policy
```
will this approach work?

recommend/report_html.go
vishnusomank: yes, i updated the code because it was showing blank for action and severity as 0 if there is no policy-level action/severity. 
should we show multiple severity and action in the table?

nyrahul (changes requested):

Issue #190, opened 2022-10-01
handling k8s port-forwarding internally 


nyrahul:
Currently, when someone uses `karmor log` or `karmor discover` or `karmor summary`, they get thrown an error message to do port-forwarding.

Wondering if we should do port-forwarding internally in kubearmor-client without throwing an error.

Cases to handle:
- [ ] if the port-forward already exist to the 32767 (kubearmor-relay) or 9089 (discovery-engine) ... just proceed as usual
- [ ] if port-forward does not exist then create a port-forward routine with unused port (between a range, 32768-32900)
- [ ] handle it for kubearmor-relay port
- [ ] handling it for discovery-engine port

PR #196, opened 2022-10-17
chore(log): handle port forward internally

rajaSahil:
Signed-off-by: rajaSahil <sahilraja242@gmail.com>

- Handle Port forward internally for log, summary and discover

Fix for #190

### DIFF
--- a/log/log.go
+++ b/log/log.go
@@ -11,8 +11,12 @@ import (
 	"os"
 	"os/signal"
 	"regexp"
+	"strconv"
 	"syscall"
 	"time"
+
+	"github.com/kubearmor/kubearmor-client/k8s"
+	"github.com/kubearmor/kubearmor-client/utils"
 )
 
 type regexType *regexp.Regexp
@@ -51,6 +55,7 @@ type Options struct {
 var StopChan chan struct{}
 var sigChan chan os.Signal
 var unblockSignal = false
+var matchLabels = map[string]string{"kubearmor-app": "kubearmor-relay"}
 
 // GetOSSigChannel Function
 func GetOSSigChannel() chan os.Signal {
@@ -109,7 +114,7 @@ func closeStopChan() {
 }
 
 // StartObserver Function
-func StartObserver(o Options) error {
+func StartObserver(c *k8s.Client, o Options) error {
 	gRPC := "localhost:32767"
 
 	if o.GRPC != "" {
@@ -117,8 +122,6 @@ func StartObserver(o Options) error {
 	} else if val, ok := os.LookupEnv("KUBEARMOR_SERVICE"); ok {
 		gRPC = val
 	}
-
-	fmt.Fprintln(os.Stderr, "gRPC server: "+gRPC)
 
 	if o.MsgPath == "none" && o.LogPath == "none" {
 		flag.PrintDefaults()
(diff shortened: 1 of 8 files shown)


log/log.go
```diff
@@ -132,9 +135,17 @@ func StartObserver(o Options) error {
 
 	// create a client
 	logClient := NewClient(gRPC, o.MsgPath, o.LogPath, o.LogFilter, o.Limit)
-	if logClient == nil {
-		return errors.New("failed to connect to the gRPC server\nPossible troubleshooting:\n- Check if Kubearmor is running\n- Create a portforward to KubeArmor relay service using\n\t\033[1mkubectl -n kube-system port-forward service/kubearmor --address 0.0.0.0 --address :: 32767:32767\033[0m\n- Configure grpc server information using\n\t\033[1mkarmor log --grpc <info>\033[0m")
+	for logClient == nil {
+		pf, err := utils.InitiatePortForward(c, 32767, 32767, matchLabels)
```
Ankurk99, line 139: Can we assign the port no. to a variable and then pass that here?

utils/portforward.go
```diff
@@ -0,0 +1,141 @@
...
+	"time"
+
+	"github.com/kubearmor/kubearmor-client/k8s"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/tools/portforward"
+	"k8s.io/client-go/transport/spdy"
+)
+
+// Port forward details for a pod
+type PortForwardOpt struct {
```
Ankurk99, line 22: Can you also please address the linting issues?

utils/portforward.go
```diff
@@ -0,0 +1,141 @@
...
+	if err := pf.getPodName(c); err != nil {
+		return err
+	}
+
+	// local port
+	pf.LocalPort = pf.GetLocalPort()
+
+	err := k8sPortForward(c, *pf)
+	if err != nil {
+		fmt.Printf("could not do kubearmor portforward, Error=%s", err.Error())
```
Ankurk99, line 58: Would fmt.Error() make more sense?

utils/portforward.go
```diff
@@ -0,0 +1,141 @@
...
+	}
+	return nil
+
+}
+
+// k8s port forward
+func k8sPortForward(c *k8s.Client, pf PortForwardOpt) error {
+	roundTripper, upgrader, err := spdy.RoundTripperFor(c.Config)
+	if err != nil {
+		fmt.Printf("unable to spdy.RoundTripperFor error=%s", err.Error())
```
Ankurk99, line 69: It will display "spdy.RoundTripperFor" and then the error. Can we make it more clear?

utils/portforward.go
```diff
@@ -0,0 +1,141 @@
...
+	podList, err := c.K8sClientset.CoreV1().Pods(pf.Namespace).List(context.Background(), metav1.ListOptions{
+		LabelSelector: metav1.FormatLabelSelector(&labelSelector),
+	})
+
+	if err != nil {
+		return err
+	}
+	if len(podList.Items) == 0 {
+		fmt.Println("Kubearmor Pod not found")
+		err = fmt.Errorf("kubearmor pod not found")
```
Ankurk99, line 119: Why use it twice? fmt.Errorf() is enough

Ankurk99 (changes requested):

log/log.go
nyrahul, line 139: With this logic, if someone tries to execute `karmor log` from multiple terminals ... It will still fail since the local port 32767 will be busy.

We should keep this logic in a for loop for maybe 10/20 ports starting 32767.

```
// Pseudo code
NewClient() // default New Client
for lport = 32767; lport < 32767 + 20 ; lport++  {
     err = InitiatePortForward(c, serverPort, lport, matchLabels)
     if err == nil {
          NewClient()
          break
     }
}
```

log/log.go
```diff
@@ -132,9 +135,17 @@ func StartObserver(o Options) error {
...
+		if err != nil {
+			return err
+		}
+		gRPC = "localhost:" + strconv.Itoa(pf.LocalPort)
+		logClient = NewClient(gRPC, o.MsgPath, o.LogPath, o.LogFilter, o.Limit)
 	}
+
+	fmt.Fprintln(os.Stderr, "gRPC server: "+gRPC)
+
```
nyrahul, line 148:
```suggestion
```

utils/portforward.go
```diff
@@ -0,0 +1,141 @@
+package utils
```
nyrahul, line 1: Rather than copying the file, can this be imported from kubearmor?

nyrahul (changes requested):
The current logic is handled only for `karmor log` .. the same logic needs to be added for `karmor summary` & `karmor discover` which connects to the discovery-engine.

log/log.go
nyrahul: Question is, if it failed outside of for loop .. it will also fail continuously within the for loop as well since the same ports are used.

nyrahul (changes requested):

utils/portforward.go
nyrahul: Makes sense. Then lets remove it from kubearmor code and import there.

nyrahul (approved):

PR #208, opened 2022-11-16
[fix] karmor ginkgo test

vishnusomank:
- updated policy data in res/out folder for  ginkgo test
- updated code with new comparison logic

Signed-off-by: Vishnu Soman <vishnu@accuknox.com>

### DIFF
--- a/tests/recommend/recommend_test.go
+++ b/tests/recommend/recommend_test.go
@@ -4,12 +4,14 @@
 package recommend_test
 
 import (
+	"encoding/json"
 	"fmt"
 	"io/fs"
 	"os"
 	"path/filepath"
 
 	. "github.com/onsi/ginkgo/v2"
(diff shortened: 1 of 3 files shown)


Ankurk99 (approved):
LGTM!

tests/recommend/recommend_test.go
```diff
@@ -23,15 +25,42 @@ var err error
...
+		fmt.Println(err)
+		return false
+	}
+	err = json.Unmarshal(policyRulesJSON, &pol2)
+	if err != nil {
+		fmt.Println(err)
+		return false
+	}
+
+	if !cmp.Equal(pol1, pol2) {
```
nyrahul, line 59: This `cmp.Equal` will fail since the discovered yaml won't exactly be in same order..

nyrahul (changes requested):

PR #210, opened 2022-11-17
Adding support to display k8s Network policy in discover

seswarrajan:

### DIFF
--- a/discover/discover.go
+++ b/discover/discover.go
@@ -79,22 +79,17 @@ func ConvertPolicy(c *k8s.Client, o Options) error {
 		return errors.New("could not connect to the server. Possible troubleshooting:\n- Check if discovery engine is running\n- Create a portforward to discovery engine service using\n\t\033[1mkubectl port-forward -n explorer service/knoxautopolicy --address 0.0.0.0 --address :: 9089:9089\033[0m\n- Configure grpc server information using\n\t\033[1mkarmor log --grpc <info>\033[0m")
 	}
 
-	if o.Policy == "network" {
-		policy := types.CiliumNetworkPolicy{}
-
-		ciliumpolicy := []types.CiliumNetworkPolicy{}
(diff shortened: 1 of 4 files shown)


discover/discover.go
```diff
@@ -149,11 +182,13 @@ func ConvertPolicy(c *k8s.Client, o Options) error {
 // Policy discovers Cilium or KubeArmor policies
 func Policy(c *k8s.Client, o Options) error {
 	if o.Policy == "cilium" {
-		o.Policy = "network"
+		o.Policy = "network-cilium"
```
nyrahul, line 185: Can we use the `kind` in place of using our own term?
```suggestion
		o.Policy = "CiliumNetworkPolicy"
```

discover/discover.go
```diff
@@ -149,11 +182,13 @@ func ConvertPolicy(c *k8s.Client, o Options) error {
...
 	} else if o.Policy == "kubearmor" {
 		o.Policy = "system"
+	} else if o.Policy == "k8snetpol" {
+		o.Policy = "network-generic"
```
nyrahul, line 189: Can we use the `kind` in place of using our own term?
```suggestion
		o.Policy = "NetworkPolicy"
```

discover/discover.go
```diff
@@ -149,11 +182,13 @@ func ConvertPolicy(c *k8s.Client, o Options) error {
 // Policy discovers Cilium or KubeArmor policies
 func Policy(c *k8s.Client, o Options) error {
 	if o.Policy == "cilium" {
-		o.Policy = "network"
+		o.Policy = "network-cilium"
 	} else if o.Policy == "kubearmor" {
 		o.Policy = "system"
```
nyrahul, line 187: Use `KubearmorSecurityPolicy`

nyrahul (changes requested):
This is superb. I was going to ask for this today and we already have it. Thanks for taking this up proactively. Few comments inline.

PR #216, opened 2022-11-28
update client to use single kubearmor controller

Ankurk99:
With https://github.com/kubearmor/KubeArmor/pull/1001, we can remove the dependency on kubearmor-host-policy and kubearmor-policy controllers
Tested locally


achrefbensaad (approved):
LGTM!

PR #221, opened 2022-12-09
add FULL enforcement if active LSM is bpf

kranurag7:
Signed-off-by: Anurag <contact.anurag7@gmail.com>

- set enforcement to full when active lsm is bpflsm
- tested on RHEL 9 sandbox


daemon1024 (approved):

daemon1024:
Thanks for the PR. It was much needed :D

PR #227, opened 2023-01-01
add verbose flag for karmor probe

### LINKED ISSUE #226: Request for new flag while `karmor probe`
When using `karmor probe` the node information takes the most time to load. While the node value is very important and is very helpful during debugging, it is something that one might not want to `stdout` every time one runs `karmor probe`. When you're running a large cluster, then getting to `Armored up pods` is taking a long time. 
One should get to the `Armored up pods` directly and if something is breaking then they should take a look at node related information. 

**Proposed Solution:**

- We can introduce a flag that will only print `armored up pods` section (preferred) OR 
- We can introduce a `--node=<node-name>` flag that will only print the respective node specific information and if set to `false` it will not print any node related information.

kranurag7:
Signed-off-by: Anurag <contact.anurag7@gmail.com>

Fixes #226 

- Refactored the output of `karmor probe`.
- The usual loading time is 1/3rd of the existing one. 
- I tested this on a 3 node GKE cluster and results are following. 
```bash
# truncated
$ time ./kubearmor-client probe 
./kubearmor-client probe  0.09s user 0.01s system 1% cpu 6.258 total
$ time ./kubearmor-client probe -v 
./kubearmor-client probe -v  0.16s user 0.05s system 1% cpu 18.309 total
```
- The new output looks something like this. 
```bash
$ ./kubearmor-client probe

Found KubeArmor running in Kubernetes

Daemonset :
 	kubearmor 	Desired: 3	Ready: 3	Available: 3	
Armored Up pods :
+-----------+------+--------+
| NAMESPACE | NAME | POLICY |
+-----------+------+--------+
| default   | pod  |        |
+-----------+------+--------+
```

### DIFF
--- a/cmd/probe.go
+++ b/cmd/probe.go
@@ -25,16 +25,9 @@ If KubeArmor is running, It probes which environment KubeArmor is running on (e.
 the supported KubeArmor features in the environment, the pods being handled by KubeArmor and the policies running on each of these pods`,
 	RunE: func(cmd *cobra.Command, args []string) error {
 		if !verbose {
-			err := probe.PrintAnnotatedPods(client, probeInstallOptions)
-			if err != nil {
-				return err
-			}
-		} else {
-			if err := probe.PrintProbeResult(client, probeInstallOptions); err != nil {
-				return err
-			}
+			return probe.PrintAnnotatedPods(client, probeInstallOptions)
 		}
-		return nil
+		return probe.PrintProbeResult(client, probeInstallOptions); err != nil {
 	},
 }
 
--- a/cmd/probe.go
+++ b/cmd/probe.go
@@ -8,6 +8,8 @@ import (
 	"github.com/spf13/cobra"
 )
 
+var verbose bool
+
 var probeInstallOptions probe.Options
 
 // probeCmd represents the get command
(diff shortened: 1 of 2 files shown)


probe/probe.go
```diff
@@ -70,6 +70,17 @@ func probeDaemonUninstaller(c *k8s.Client, o Options) error {
 	return nil
 }
 
+// only annotated pods and corresponding policies
```
nyrahul, line 73:
```suggestion
// PrintAnnotatedPods only annotated pods and corresponding policies
```

cmd/probe.go
```diff
@@ -22,16 +24,23 @@ and what KubeArmor features will be supported e.g: observability, enforcement, e
...
+			err := probe.PrintAnnotatedPods(client, probeInstallOptions)
+			if err != nil {
+				return err
+			}
+		} else {
+			if err := probe.PrintProbeResult(client, probeInstallOptions); err != nil {
+				return err
+			}
 		}
 		return nil
```
nyrahul, line 37:
```suggestion
		if !verbose {
			return probe.PrintAnnotatedPods(client, probeInstallOptions)
		}
		return probe.PrintProbeResult(client, probeInstallOptions); err != nil {
```

nyrahul (changes requested):

yasin-cs-ko-ak:
@kranurag7 Kindly review the suggested changes.
cc: @nyrahul

kranurag7:
This proposal is not planned as of now (as per our discussion in community meetings). Closing this. `karmor probe` does the job.

PR #236, opened 2023-01-04
Fixed karmor probe to show correct policies applied based on MatchLabels 

### LINKED ISSUE #205: Invalid apply of ksp - problem with using selector labels
## Bug Report

Selector labels construct in the policy tells the kubearmor which pods the corresponding policy has to be applied on.

Consider [wordpress-mysql example](https://github.com/kubearmor/KubeArmor/tree/main/examples/wordpress-mysql):

kubectl apply -f policy.yaml
```yaml
apiVersion: security.kubearmor.com/v1
kind: KubeArmorPolicy
metadata:
  name: ksp-wordpress-block-process
  namespace: wordpress-mysql
spec:
  severity: 3
  selector:
    matchLabels:
      app: wordpress
      env: production
  process:
    matchPaths:
    - path: /usr/bin/apt
    - path: /usr/bin/apt-get
  action:
    Block
```

Note that the wordpress deployment does not have the label `env: production` but still the policy is applied as shown by `karmor probe`.

```
Armored Up pods : 
+-----------------+---------------------------------+-----------------------------+
|    NAMESPACE    |              NAME               |           POLICY            |
+-----------------+---------------------------------+-----------------------------+
| wordpress-mysql | mysql-58cdf6ccf-z9hpc           |                             |
+                 +---------------------------------+-----------------------------+
|                 | wordpress-bf95888cb-56x8j       | ksp-wordpress-block-process |
+-----------------+---------------------------------+-----------------------------+
```

**To Reproduce**

1. apply the policy.yaml
2. check `karmor probe`

**Expected behavior**

`karmor probe` should not show policy applied in the context of wordpress pods.

PrimalPimmy:
Fixes #205 

## The Problem before

The issue with karmor probe was how labels were handled, where if even one `MatchLabel` from the policy matched with the current Pod Label, it would hold true, i.e using Logical OR instead of logical AND.

## The Solution

I used the package https://github.com/deckarep/golang-set, and the logic is that there are two sets (like python sets): Pod Labels and the MatchLabel mentioned in the ksp policy, and now `karmor probe` will show that the policy is only applied when all labels mentioned in `MatchLabels`  is there in the actual Pod Label. 
I found this easier than to manually implement to compare two different slices.

The time complexity of above approach is linear time. The time complexity of issubset itself is O(n), where n is length of a set we are checking if subset, in this case s2. There is also conversion from slice to set that is also linear time.

image

### DIFF
--- a/probe/probe.go
+++ b/probe/probe.go
@@ -605,7 +605,6 @@ func getAnnotatedPods(c *k8s.Client) error {
 			s1 := sliceToSet(labels)
 			for policyKey, policyValue := range policyMap {
 				s2 := sliceToSet(policyValue)
-				// log.Print(policyValue)
 				if s2.IsSubset(s1) {
 					if checkIfDataAlreadyContainsPodName(data, armoredPod.Name, policyKey) {
 						continue
--- a/probe/probe.go
+++ b/probe/probe.go
@@ -17,6 +17,7 @@ import (
 	"strings"
 	"time"
 
+	mapset "github.com/deckarep/golang-set/v2"
 	"github.com/fatih/color"
 	jsoniter "github.com/json-iterator/go"
 	tp "github.com/kubearmor/KubeArmor/KubeArmor/types"
@@ -571,6 +572,14 @@ func probeSystemdMode() error {
 	return nil
 }
 
+func getAnnotatedPodLabels(m map[string]string, a []string) []string {
+	for key, value := range m {
+		a =append(a, key+":"+value)
+	}
+
+	return a
+}
+
 func getAnnotatedPods(c *k8s.Client) error {
 	// Annotated Pods Description
 	var data [][]string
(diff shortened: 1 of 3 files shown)


probe/probe.go
```diff
@@ -590,18 +599,18 @@ func getAnnotatedPods(c *k8s.Client) error {
...
-			for key, value := range armoredPod.Labels {
-				for policyKey, policyValue := range policyMap {
-					if slices.Contains(policyValue, key+":"+value) {
-						if checkIfDataAlreadyContainsPodName(data, armoredPod.Name, policyKey) {
-							continue
-						} else {
-							data = append(data, []string{armoredPod.Namespace, armoredPod.Name, policyKey})
-						}
-
+			labels = getAnnotatedPodLabels(armoredPod.Labels, labels)
```
nyrahul, line 604: Can we directly return a mapset from the getAnnotatedPodLabels()? Rather than returning first a string slice (labels( and then converting the slice to mapset (s1).

nyrahul (changes requested):

nyrahul (approved):

PR #238, opened 2023-01-09
Usability Improvements in karmor profile command

PrimalPimmy:
Added the ability for the user to change the max number of rows per page, by default it will show 30 rows.
rows


Ankurk99:
Hey! Can you please add flag for `--gRPC`? Also there are issues with the way the cli responds to font and terminal size. But this I think can be opened as a separate issue and be addressed in subsequent PR.

cmd/profile.go
```diff
@@ -23,7 +22,7 @@ var profilecmd = &cobra.Command{
 
 func init() {
 	rootCmd.AddCommand(profilecmd)
-
+	profilecmd.Flags().StringVar(&profileOptions.GRPC, "gRPC", "", "use gRPC")
```
seswarrajan, line 25: Can we have default of 32767 port ?

nyrahul:
Hey @PrimalPimmy , can you fix the go-fmt issues?

PrimalPimmy:
Yes @nyrahul I will. Also I reviewed with @Ankurk99 and found some usability bugs so I'm gonna be adding some new fixes before this can be merged. Will let you know.

cmd/profile.go
PrimalPimmy: Yeah but for some reason when I try to run karmor log or profile, the default port is different for me if I don't mention `--gRPC` (when cluster is hosted in VM and accessed from outside).

Now, do we want the default scenario to be when cluster and karmor is on host itself? Then it can be default.

PrimalPimmy:
I have made some major quality changes to profiling, these changes are:
cc: @Ankurk99 
- Added a gRPC flag
- Finally fixed the artefacts that would be caused if the terminal was resized
- Flexible width of the table so changing font size won't be an issue anymore

PR #245, opened 2023-01-18
updated error message for failing to find pods before port forward

TheRealSibasishBehera:
 
  * while doing port-forward , if pod is not found , the error message is always 
           `Error: kubearmor pod not found`
           
           
Now the new error message also specifies the pod that was supposed to be found
For example in case of `karmor summary` if the client fails to find to find the pod
  
`Error: kubearmor discovery-engine pod not found`

### DIFF
--- a/discover/discover.go
+++ b/discover/discover.go
@@ -42,6 +42,7 @@ var port int64 = 9089
 // ConvertPolicy converts the knoxautopolicies to KubeArmor and Cilium policies
 func ConvertPolicy(c *k8s.Client, o Options) error {
 	gRPC := ""
+	targetPodName := "discovery-engine"
 
 	if o.GRPC != "" {
 		gRPC = o.GRPC
@@ -49,7 +50,7 @@ func ConvertPolicy(c *k8s.Client, o Options) error {
 		if val, ok := os.LookupEnv("DISCOVERY_SERVICE"); ok {
 			gRPC = val
 		} else {
-			pf, err := utils.InitiatePortForward(c, port, port, matchLabels)
+			pf, err := utils.InitiatePortForward(c, port, port, matchLabels, targetPodName)
 			if err != nil {
 				return err
 			}
--- a/discover/discover.go
+++ b/discover/discover.go
@@ -42,7 +42,7 @@ var port int64 = 9089
 // ConvertPolicy converts the knoxautopolicies to KubeArmor and Cilium policies
 func ConvertPolicy(c *k8s.Client, o Options) error {
 	gRPC := ""
-	targetPodName := "discovery-engine"
+	targetSvc := "discovery-engine"
 
 	if o.GRPC != "" {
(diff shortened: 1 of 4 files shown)


discover/discover.go
```diff
@@ -42,14 +42,15 @@ var port int64 = 9089
 // ConvertPolicy converts the knoxautopolicies to KubeArmor and Cilium policies
 func ConvertPolicy(c *k8s.Client, o Options) error {
 	gRPC := ""
+	targetPodName := "discovery-engine"
```
nyrahul, line 45: lets call this targetSvc
```suggestion
	targetSvc := "discovery-engine"
```

utils/portforward.go
```diff
@@ -22,20 +22,22 @@ import (
...
+	RemotePort    int64
+	MatchLabels   map[string]string
+	Namespace     string
+	PodName       string
+	TargetPodName string
 }
 
 // InitiatePortForward : Initiate port forwarding
-func InitiatePortForward(c *k8s.Client, localPort int64, remotePort int64, matchLabels map[string]string) (PortForwardOpt, error) {
+func InitiatePortForward(c *k8s.Client, localPort int64, remotePort int64, matchLabels map[string]string, targetPodName string) (PortForwardOpt, error) {
```
nyrahul, line 34:
```suggestion
func InitiatePortForward(c *k8s.Client, localPort int64, remotePort int64, matchLabels map[string]string, targetSvc string) (PortForwardOpt, error) {
```

utils/portforward.go
```diff
@@ -22,20 +22,22 @@ import (
...
 	pf := PortForwardOpt{
-		LocalPort:   localPort,
-		RemotePort:  remotePort,
-		Namespace:   "",
-		MatchLabels: matchLabels,
+		LocalPort:     localPort,
+		RemotePort:    remotePort,
+		Namespace:     "",
+		MatchLabels:   matchLabels,
+		TargetPodName: targetPodName,
```
nyrahul, line 40:
```suggestion
		TargetSvc: targetSvc,
```

utils/portforward.go
```diff
@@ -118,7 +120,7 @@ func (pf *PortForwardOpt) getPodName(c *k8s.Client) error {
 		return err
 	}
 	if len(podList.Items) == 0 {
-		return errors.New("kubearmor pod not found")
+		return fmt.Errorf("kubearmor %s pod not found", pf.TargetPodName)
```
nyrahul, line 123:
```suggestion
		return fmt.Errorf("%s svc not found", pf.TargetSvc)
```

nyrahul (changes requested):

discover/discover.go
TheRealSibasishBehera: Yeah , will change `PodName` to `Svc`

nyrahul (approved):

PR #247, opened 2023-01-24
updated localPort value to not bind with with any port to be used by …

TheRealSibasishBehera:

Fixes [KubeArmor-#1051](https://github.com/kubearmor/KubeArmor/issues/1051)  

 * previously karmor used to bind to port 32767 , if `karmor log` was used before all KubeArmor were running during `karmor install` . This caused `kubearmor` pod failing to run , as it requires to use 32767

     
 * Now karmor would not bind to 32767 , instead keep looking for other local ports  > 32767

### DIFF
--- a/utils/portforward.go
+++ b/utils/portforward.go
@@ -129,8 +129,6 @@ func (pf *PortForwardOpt) getPodName(c *k8s.Client) error {
 func (pf *PortForwardOpt) getLocalPort() (int64, error) {
 
 	for {
-		port := pf.LocalPort
-
 		port, err := getRandomPort()
 		if err != nil {
 			return port, err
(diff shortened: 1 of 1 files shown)


utils/portforward.go
```diff
@@ -127,9 +127,14 @@ func (pf *PortForwardOpt) getPodName(c *k8s.Client) error {
 
 // Returns the local port for the port forwarder
 func (pf *PortForwardOpt) getLocalPort() (int64, error) {
-	port := pf.LocalPort
 
 	for {
+		port := pf.LocalPort
```
Ankurk99, line 132: not required as it's already defined in the next step

Ankurk99 (approved):
LGTM! Just a nit suggested inline

Ankurk99:
Thanks for the PR @TheRealSibasishBehera :rocket:

PR #251, opened 2023-01-31
Added check for update in any mandatory release of karmor CLI

### LINKED ISSUE #249: Check for karmor client updates if mandatory updates required
Currently Karmor client does not inform the user if there's any major update to the client. This might lead to breaking functionalities.

## Suggestion
Do a check when starting karmor cli if there is a mandatory release mentioned in the release notes of karmor client. Then informs the user that they need to update the client in the CLI.

TheRealSibasishBehera:

Fixes #249 

 the function checkForUpdates first finds the latest release that contains the keyword "mandatory" in its release notes 
 compared it with current version from `selfupdate.GitSummary`
 If diffrent informs the user

### DIFF
--- a/checks/updates.go
+++ b/checks/updates.go
@@ -1,0 +1,53 @@
+package checks
+
+import (
+	"context"
+	"fmt"
+	"github.com/fatih/color"
+	"github.com/google/go-github/github"
+	"github.com/kubearmor/kubearmor-client/selfupdate"
+	"strings"
+)
+
+func CheckForUpdates() error {
+	client := github.NewClient(nil)
+
+	// Fetch all releases from the GitHub API
+	releases, _, err := client.Repositories.ListReleases(context.Background(),
+		"kubearmor",
+		"kubearmor-client",
+		&github.ListOptions{
+			Page:    1,
+			PerPage: 100,
+		},
+	)
+	if err != nil {
+		fmt.Println("Error fetching releases:", err)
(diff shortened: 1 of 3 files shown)


PrimalPimmy:
Hey @TheRealSibasishBehera . Great work on this 🚀. 
Could you show an image/gif that this feature is working? If tested.

TheRealSibasishBehera:
Screenshot from 2023-01-31 16-21-01

the first stdout `The client is up to data` is from `CheckForUpdates()`

For this feature to identify major updates , the release notes must have a keyword `mandatory` , i dont think this keyword is used in release notes till now .

On further releases we can add the `mandatory` keyword in case of a major update

If its alright i can add a unit test with a mock `*RepositoryRelease` , which have that keyword and test it.

PrimalPimmy:
@TheRealSibasishBehera i don't think a test would be necessary for this feature. As long as the keyboard isn't case sensitive. If it is case sensitive, could you make it "MANDATORY" all caps?

TheRealSibasishBehera:
@PrimalPimmy I have updated the code , please take a look

nyrahul (changes requested):
This is an incorrect PR. This PR implements the logic in its own binary. The aim of the task was to check whether a mandatory release exists in the github when someone calls `karmor version` and highlight that a mandatory upgrade is pending.

PR #266, opened 2023-02-10
printing string output as valid json

Prateeknandle:
Signed-off-by: Prateeknandle <prateeknandle@gmail.com>
The current output was not a valid json output, now we can use the output directly as a valid json


Prateeknandle:
covered in https://github.com/kubearmor/kubearmor-client/pull/268

PR #268, opened 2023-02-13
refractor summary api

Prateeknandle:
Signed-off-by: Prateeknandle <prateeknandle@gmail.com>
refactoring summary api:
changes : 
- returning the summary output when `GetSummary()` is called
- printing the output is now done by `Summary()` func, when comand is used

after this we can use summary api to get the `karmor summary` command output rather than executing command in code.

### DIFF
--- a/summary/summary.go
+++ b/summary/summary.go
@@ -39,8 +39,9 @@ type Options struct {
 	Aggregation   bool
 }
 
-// Summary : Get summary on pods
-func Summary(c *k8s.Client, o Options) error {
+// GetSummary on pods
+func GetSummary(c *k8s.Client, o Options) ([]string, error) {
+	var str []string
 	gRPC := ""
 	targetSvc := "discovery-engine"
 
@@ -52,7 +53,7 @@ func Summary(c *k8s.Client, o Options) error {
 		} else {
 			pf, err := utils.InitiatePortForward(c, port, port, matchLabels, targetSvc)
 			if err != nil {
-				return err
+				return nil, err
 			}
 			gRPC = "localhost:" + strconv.FormatInt(pf.LocalPort, 10)
 		}
@@ -70,7 +71,7 @@ func Summary(c *k8s.Client, o Options) error {
 	// create a client
 	conn, err := grpc.Dial(gRPC, grpc.WithTransportCredentials(insecure.NewCredentials()))
 	if err != nil {
-		return errors.New("could not connect to the server. Possible troubleshooting:\n- Check if discovery engine is running\n- Create a portforward to discovery engine service using\n\t\033[1mkubectl port-forward -n explorer service/knoxautopolicy --address 0.0.0.0 --address :: 9089:9089\033[0m\n[0m")
+		return nil, errors.New("could not connect to the server. Possible troubleshooting:\n- Check if discovery engine is running\n- kubectl get po -n accuknox-agents")
 	}
 	defer conn.Close()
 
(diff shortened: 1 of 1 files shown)


summary/summary.go
```diff
@@ -83,15 +84,15 @@ func Summary(c *k8s.Client, o Options) error {
 			Aggregate: o.Aggregation,
 		})
 		if err != nil {
-			return err
+			return nil, err
 		}
 		DisplaySummaryOutput(sumResp, o.RevDNSLookup, o.Type)
```
rksharma95, line 90: there is one bug already here
if summary is to be filtered based on given pod name and output format is json then we'll not get desired output.
can you handle this case too. and we need to return the result too if the format is json.

summary/summary.go
```diff
@@ -104,19 +105,37 @@ func Summary(c *k8s.Client, o Options) error {
...
+			sumstr := ""
 			if o.Output == "json" {
 				arr, _ := json.MarshalIndent(sumResp, "", "    ")
-				str = fmt.Sprintf("%s\n", string(arr))
-				fmt.Printf("%s", str)
+				sumstr = fmt.Sprintf("%s\n", string(arr))
+				str = append(str, sumstr)
 			}
 		}
+		if o.Output == "json" {
```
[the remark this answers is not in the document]
Prateeknandle: updated, please check

summary/summary.go
Prateeknandle: @rksharma95 updated, please check

summary/summary.go
```diff
@@ -70,7 +71,7 @@ func Summary(c *k8s.Client, o Options) error {
 	// create a client
 	conn, err := grpc.Dial(gRPC, grpc.WithTransportCredentials(insecure.NewCredentials()))
 	if err != nil {
-		return errors.New("could not connect to the server. Possible troubleshooting:\n- Check if discovery engine is running\n- Create a portforward to discovery engine service using\n\t\033[1mkubectl port-forward -n explorer service/knoxautopolicy --address 0.0.0.0 --address :: 9089:9089\033[0m\n[0m")
+		return nil, errors.New("could not connect to the server. Possible troubleshooting:\n- Check if discovery engine is running\n- Create a portforward to discovery engine service using\n\t\033[1mkubectl port-forward -n explorer service/knoxautopolicy --address 0.0.0.0 --address :: 9089:9089\033[0m\n[0m")
```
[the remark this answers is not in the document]
Prateeknandle: @yasin-cs-ko-ak i'm not getting what's the problem here and how to resolve it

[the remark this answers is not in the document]
rksharma95: the discovery-engine is no longer deployed in `explorer` namespace by default.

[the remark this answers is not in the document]
Prateeknandle: updated please check

summary/summary.go
```diff
@@ -70,7 +71,7 @@ func Summary(c *k8s.Client, o Options) error {
 	// create a client
 	conn, err := grpc.Dial(gRPC, grpc.WithTransportCredentials(insecure.NewCredentials()))
 	if err != nil {
-		return errors.New("could not connect to the server. Possible troubleshooting:\n- Check if discovery engine is running\n- Create a portforward to discovery engine service using\n\t\033[1mkubectl port-forward -n explorer service/knoxautopolicy --address 0.0.0.0 --address :: 9089:9089\033[0m\n[0m")
+		return nil, errors.New("could not connect to the server. Possible troubleshooting:\n- Check if discovery engine is running\n- Create a portforward to discovery engine service using\n\t\033[1mkubectl port-forward -n <namespace> service/knoxautopolicy --address 0.0.0.0 --address :: 9089:9089\033[0m\n[0m")
```
yasin-cs-ko-ak, line 74: Since port-forwarding is happening internally. We don't have to mention port-forward explicitly.
so the error message could be like this 
```console
"could not connect to the server. Possible troubleshooting:\n- Check if discovery engine is running\n- kubectl get po -n accuknox-agents"
``` 
this is enough I guess

rksharma95 (approved):
LGTM :+1: PTAL at comment by @yasin-cs-ko-ak

summary/summary.go
Prateeknandle: ok

Prateeknandle: @yasin-cs-ko-ak updated the err msg

yasin-cs-ko-ak (approved):
Good Job!
LGTM

yasin-cs-ko-ak:
@Prateeknandle If you can attach a screenshot of a JSON output, that would be better.

Prateeknandle:
screenshot was not possible because summary is big, therefore pasting the output with command

rksharma95:
@yasin-cs-ko-ak is it okay to merge this PR now?

rksharma95:
@Prateeknandle @yasin-cs-ko-ak IMO we're good to merge it. 
// cc: @seswarrajan

PR #274, opened 2023-03-03
enhancement(karmor recommend) : hardening policies consolidation

yasin-cs-ko-ak:
This PR contains enhancements for hardening policies (karmor recommend)
- Merging recommended policies into single policy using `--merge` flag

### DIFF
--- a/cmd/recommend.go
+++ b/cmd/recommend.go
@@ -50,4 +50,6 @@ func init() {
 	recommendCmd.Flags().StringVarP(&recommendOptions.ReportFile, "report", "r", "report.txt", "report file")
 	recommendCmd.Flags().StringSliceVarP(&recommendOptions.Tags, "tag", "t", []string{}, "tags (comma-separated) to apply. Eg. PCI-DSS, MITRE")
 	recommendCmd.Flags().StringVarP(&recommendOptions.Config, "config", "c", recommend.UserHome()+"/.docker/config.json", "absolute path to image registry configuration file")
+	recommendCmd.Flags().BoolVarP(&recommendOptions.Merge, "merge", "m", false, "Merge recommended hardening policies")
+
 }
--- a/recommend/recommend.go
+++ b/recommend/recommend.go
@@ -7,13 +7,17 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"hash/fnv"
 	"os"
 	"path/filepath"
+	"sort"
+	"strconv"
 	"strings"
 
 	"github.com/fatih/color"
 	"github.com/kubearmor/kubearmor-client/k8s"
 	log "github.com/sirupsen/logrus"
+	"gopkg.in/yaml.v2"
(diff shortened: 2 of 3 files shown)


cmd/recommend.go
```diff
@@ -50,4 +50,6 @@ func init() {
 	recommendCmd.Flags().StringVarP(&recommendOptions.ReportFile, "report", "r", "report.txt", "report file")
 	recommendCmd.Flags().StringSliceVarP(&recommendOptions.Tags, "tag", "t", []string{}, "tags (comma-separated) to apply. Eg. PCI-DSS, MITRE")
 	recommendCmd.Flags().StringVarP(&recommendOptions.Config, "config", "c", recommend.UserHome()+"/.docker/config.json", "absolute path to image registry configuration file")
+	recommendCmd.Flags().StringVarP(&recommendOptions.Merge, "merge", "m", "merged-policies.yaml", "Merge hardening policies")
```
vishnusomank, line 53: @yasin-cs-ko-ak can we have the merge flag should be boolean with default value as false

recommend/recommend.go
```diff
@@ -39,6 +44,106 @@ type Deployment struct {
 	Images    []string
 }
 
+// ======================== //
+// == Knox System Policy == //
+// ======================== //
+
+// KnoxFromSource Structure
+type KnoxFromSource struct {
```
vishnusomank, line 52: can we import this from kubearmor and use? rather than redefining them?

recommend/recommend.go
```diff
@@ -106,6 +211,10 @@ func Recommend(c *k8s.Client, o Options) error {
 		return err
 	}
 
+	if o.Merge == "merge" {
```
vishnusomank, line 214: this will throw error if the user is making use of `karmor recommend` for the first time since there will be no folder to look for policies.

cmd/recommend.go
yasin-cs-ko-ak: thank you for your thorough review and Changes are updated.

recommend/recommend.go
yasin-cs-ko-ak: thank you for your thorough review and Changes are updated.

recommend/recommend.go
daemon1024: What are the modifications exactly? Can we maybe upstream it, I don't think we should duplicate this at all unless there's an absolutely compelling reason.  Thanks

yasin-cs-ko-ak: Yes. I will remove this type and add it from the kubearmor. Kubearmor has the type structure that's need for this. (From)[https://github.com/kubearmor/KubeArmor/blob/main/pkg/KubeArmorController/api/security.kubearmor.com/v1/kubearmorhostpolicy_types.go] here.

yasin-cs-ko-ak:
Closing this PR since this feature is not needed.

PR #277, opened 2023-03-06
feat(install): add option for custom init container image

daemon1024:
Change default init container image to use stable tag cause latest can have breaking changes

Ref https://github.com/kubearmor/KubeArmor/pull/1087


DelusionalOptimist (approved):
LGTM. :rocket:

Issue #286, opened 2023-03-10
enhancement: KubeArmor recommend security report


yasin-cs-ko-ak:
## Enhancement
This is for enhancement of the kubearmor recommend security report.

The current report is showing like [this](https://githtmlpreview.netlify.app/?https://github.com/kubearmor/KubeArmor/blob/main/getting-started/hardening-report.html)

This is the enhanced security report: 

![image-20230222-100205](https://user-images.githubusercontent.com/92008211/224203190-359a9253-eef1-47c8-9c1c-196ddf86a132.png)

PR #288, opened 2023-03-10
karmor recommend security report

vishnusomank:
- updated the security report generated while using karmor recommend with new template

Fixes: https://github.com/kubearmor/kubearmor-client/issues/286

image


nyrahul (approved):
lgtm

PR #292, opened 2023-03-14
fix(logClient.go):fix panic interface conversion: interface {} is nil not string

### LINKED ISSUE #290: bug(karmor logs): karmor logs are getting panicked when we use label function
Refer below:
```shell
➜  ~ karmor logs --json -n wordpress-mysql -l app=mysql | jq .

local port to be used for port forwarding kubearmor-relay-6fddb8865b-d6v2f: 32775 
Created a gRPC client (localhost:32775)
Checked the liveness of the gRPC server
Started to watch alerts
panic: interface conversion: interface {} is nil, not string

goroutine 44 [running]:
github.com/kubearmor/kubearmor-client/log.WatchTelemetryHelper({0xc000032480, 0x232, _}, {_, _}, {{0x0, 0x0}, {0x1ea1d7a, 0x4}, {0x1ea4d05, ...}, ...})
	/home/runner/work/kubearmor-client/kubearmor-client/log/logClient.go:304 +0xd27
github.com/kubearmor/kubearmor-client/log.(*Feeder).WatchAlerts(0xc00068de00, {{0x0, 0x0}, {0x1ea1d7a, 0x4}, {0x1ea4d05, 0x6}, {0x1ea49e7, 0x6}, 0x1, ...})
	/home/runner/work/kubearmor-client/kubearmor-client/log/logClient.go:249 +0x105
created by github.com/kubearmor/kubearmor-client/log.StartObserver
	/home/runner/work/kubearmor-client/kubearmor-client/log/log.go:173 +0x5cc
```

##  To Reproduce:
1. Deploy this wordpress-mysql deployment 
```shell
kubectl apply -f https://raw.githubusercontent.com/kubearmor/KubeArmor/main/examples/wordpress-mysql/wordpress-mysql-deployment.yaml
```
2. Install karmor client
3. Deploy kubearmor in your cluster 
4. Check for logs using `karmor logs --json -n wordpress-mysql -l app=mysql | jq .` this.

xiao-jay:
fix #290 
### First
wrong code: https://github.com/kubearmor/kubearmor-client/blob/main/log/logClient.go#L304
```
func WatchTelemetryHelper(arr []byte, t string, o Options) {
	var res map[string]interface{}
	err := json.Unmarshal(arr, &res)
	if err != nil {
		return
	}

	// Filter Telemetry based on provided options
	if len(o.Selector) != 0 {
304		val := selectLabels(o, res["Labels"].(string))
		if val != nil {
			return
		}
	}

	if o.Namespace != "" {
		ns, ok := res["NamespaceName"].(string)
		if !ok {
			return
		}
		match := regexMatcher(CNamespace, ns)
		if !match {
			return
		}
	}

```
some time `len(o.Selector)` != 0  but `res["Labels"]` is nil, will lead to panic.
### Second

if `o.Selector` match one label should select it ? so i use `labels := strings.Split(res["Labels"].(string), ",")` to split multiple labels.

### DIFF
--- a/log/logClient.go
+++ b/log/logClient.go
@@ -298,10 +298,10 @@ func WatchTelemetryHelper(arr []byte, t string, o Options) {
 	if err != nil {
 		return
 	}
-
 	// Filter Telemetry based on provided options
-	if len(o.Selector) != 0 {
-		val := selectLabels(o, res["Labels"].(string))
+	if len(o.Selector) != 0 && res["Labels"] != nil {
+		labels := strings.Split(res["Labels"].(string), ",")
+		val := selectLabels(o, labels)
 		if val != nil {
 			return
 		}
@@ -465,11 +465,14 @@ func (fd *Feeder) DestroyClient() error {
 	return nil
 }
 
-func selectLabels(o Options, labels string) error {
+func selectLabels(o Options, labels []string) error {
 	for _, val := range o.Selector {
-		if val == labels {
-			return nil
+		for _, label := range labels {
+			if val == label {
+				return nil
+			}
 		}
+
 	}
 	return errors.New("Not found any flag")
 }
--- a/log/logClient_test.go
+++ b/log/logClient_test.go
@@ -50,6 +50,7 @@ func TestLogClient(t *testing.T) {
 	eventChan = make(chan EventInfo, maxEvents)
 	var o = Options{
 		EventChan: eventChan,
+		Selector:  []string{"substance=meth"},
 	}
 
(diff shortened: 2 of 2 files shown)


Shreyas220 (approved):
This PR solves two thing 

#### 1 Solves the panic issue 
Adding a check `res["Labels"] != nil` before accessing `res["Labels"].(string)`  

#### 2 Multiple label 
Before this if we had multiple labels , we would return nil by adding this we are separate the label and check   

cc: @Ankurk99 

LGTM :+1:

Issue #329, opened 2023-06-11
recommend feature remove dependency on discovery engine


nyrahul:
Currently `karmor recommend` pops up errors on cli if the discovery engine deployment is not found. This has to be fixed i.e., karmor recommend feature should work without any dependency on discovery engine.

PrimalPimmy:
@nyrahul but things like admission controller and runtime policy within `recommend` still depend on Summary data?

PR #331, opened 2023-06-12
Removed DE dependency from recommend

PrimalPimmy:
Related to issue https://github.com/kubearmor/kubearmor-client/issues/329

Due to the new admission controller implementation, the entire recommend command was dependent on the DE due to a small error management. I have fixed that. Now recommend will run even if DE is not running, it just wont display any policies that require summary data from DE if discovery engine is not in the cluster.

image

### DIFF
--- a/recommend/imageHandler.go
+++ b/recommend/imageHandler.go
@@ -529,7 +529,7 @@ func imageHandler(namespace, deployment string, labels LabelMap, imageName strin
 		}
 		err := initClientConnection(c)
 		if err != nil {
-			log.WithError(err).Error("failed to initialize DE client connection. Won't recommend admission controller policies.")
+			log.WithError(err).Info("failed to initialize DE client connection. Won't recommend admission controller policies.")
 			//return err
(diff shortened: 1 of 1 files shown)


recommend/imageHandler.go
```diff
@@ -529,13 +529,14 @@ func imageHandler(namespace, deployment string, labels LabelMap, imageName strin
...
 		err := initClientConnection(c)
 		if err != nil {
-			log.WithError(err).Error("failed to initialize client connection.")
-			return err
-		}
-		err = recommendAdmissionControllerPolicies(img)
-		if err != nil {
-			log.WithError(err).Error("failed to recommend admission controller policies.")
-			return err
+			log.WithError(err).Error("failed to initialize DE client connection. Won't recommend admission controller policies.")
```
nyrahul, line 532: Please convert this into Info/message.

nyrahul (changes requested):

nyrahul (approved):

PR #335, opened 2023-06-22
feat(install): Probe KubeArmor after installing 

### LINKED ISSUE #334: karmor install to check if the target env supports enforcement
Currently, the user explicitly has to run the `karmor probe` to check if the target env has enforcement capabilities. I got into a discussion today where the user couldn't figure why the `"Action: Audit (Block)"` was shown in the `karmor logs` output.

The user himself proposed a solution where if we can show after `karmor install` whether the target environment supports enforcement mode.

The enviroment used was CIVO with k3s cluster. I am not aware of the target distro used underneath k3s but I assume both AppArmor and BPF-LSM were not supported there.

CC: @IdealUsrname, @daemon1024

daemon1024:
Fixes #334 

Refactored Probe Code to return probe information before printing it
Leveraging the API then in Install to extract enforcement data 

Screenshot when Enforcement Works
image

When Enforcement doesn't work
image

### DIFF
--- a/k8s/env.go
+++ b/k8s/env.go
@@ -1,0 +1,87 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2023 Authors of KubeArmor
+
+package k8s
+
+import (
+	"context"
+	"strings"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"golang.org/x/mod/semver"
+)
+
+// AutoDetectEnvironment detect the environment for a given k8s context
+func AutoDetectEnvironment(c *Client) (name string) {
+	env := "none"
+
+	contextName := c.RawConfig.CurrentContext
+	clusterContext, exists := c.RawConfig.Contexts[contextName]
+	if !exists {
+		return env
+	}
+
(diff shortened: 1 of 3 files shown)


k8s/env.go
```diff
@@ -0,0 +1,84 @@
+package k8s
```
nyrahul, line 1: license header

probe/probe.go
```diff
@@ -90,10 +89,17 @@ func PrintProbeResult(c *k8s.Client, o Options) error {
 	if isKubeArmorRunning(c, o) {
 		getKubeArmorDeployments(c, o)
 		getKubeArmorContainers(c, o)
-		err := probeRunningKubeArmorNodes(c, o)
+		probeData, err := ProbeRunningKubeArmorNodes(c, o)
 		if err != nil {
 			log.Println("error occured when probing kubearmor nodes", err)
```
nyrahul, line 94: return here on err?

nyrahul (approved):
LGTM

DelusionalOptimist (commented):
Looking mostly good. Just some nits.

DelusionalOptimist (approved):

PR #336, opened 2023-06-22
chore(install): deprecate multiple kubearmor controllers/managers

daemon1024:
Ref https://github.com/kubearmor/KubeArmor/pull/1253


DelusionalOptimist (commented):
@daemon1024 we need to add some more resources for the controller. I don't know how the controllers used to work earlier but the new controller fails to work without the roles as per my understanding - https://github.com/kubearmor/KubeArmor/blob/46513d5b46e2c2ef4ebbe81a6c1767f7822bd59c/deployments/get/defaults.go#L25-L32
cc @achrefbensaad

daemon1024:
image
 Updated installation View

DelusionalOptimist (approved):
LGTM. :rocket:

Issue #297, opened 2023-03-25
karmor probe missing node related infos


kranurag7:
`karmor probe` is missing node related information on a fresh instance. 

To Reproduce:

On a fresh EC2 instance, when you run `karmor probe` then you get the following result. 
```bash
$ karmor probe

Found KubeArmor running in Kubernetes

Daemonset :
Deployments :
 	kubearmor-relay              	Desired: 1	Ready: 1	Available: 1	
 	kubearmor-policy-manager     	Desired: 1	Ready: 1	Available: 1	
 	kubearmor-host-policy-manager	Desired: 1	Ready: 1	Available: 1	
 	kubearmor-annotation-manager 	Desired: 1	Ready: 1	Available: 1	
Containers :
 	kubearmor-relay-6fddb8865b-qr4pw              	Running: 1	Image Version: kubearmor/kubearmor-relay-server:latest  	
 	kubearmor-policy-manager-cb9dbc668-4jf52      	Running: 2	Image Version: gcr.io/kubebuilder/kube-rbac-proxy:v0.8.0	
 	kubearmor-host-policy-manager-756b46d87c-645xk	Running: 2	Image Version: gcr.io/kubebuilder/kube-rbac-proxy:v0.8.0	
 	kubearmor-annotation-manager-7fc8d9b964-ddh55 	Running: 2	Image Version: gcr.io/kubebuilder/kube-rbac-proxy:v0.8.0	
 	kubearmor-dft65                               	Running: 1	Image Version: kubearmor/kubearmor:stable               	
Node 1 :
2023/03/25 09:38:04 an error occured when parsing file readObjectStart: expect { or n, but found , error found in #0 byte of ...||..., bigger context ...||...
2023/03/25 09:38:04 an error occured when printing output readObjectStart: expect { or n, but found , error found in #0 byte of ...||..., bigger context ...||...
2023/03/25 09:38:04 error occured when probing kubearmor nodes readObjectStart: expect { or n, but found , error found in #0 byte of ...||..., bigger context ...||...
Armored Up pods :
+-----------+------+--------+
| NAMESPACE | NAME | POLICY |
+-----------+------+--------+
+-----------+------+--------+
```

```bash
$ karmor version
karmor version 0.12.4 linux/amd64 BuildDate=2023-03-07T17:32:40Z
current version is the latest
kubearmor image (running) version kubearmor/kubearmor:stable
```

<details>
<summary> attached screenshot </summary>

[...]

PR #338, opened 2023-06-22
feat(install): wait for KubeArmor to create probe file before probing

daemon1024:
KubeArmor takes some time to initialise Monitor and Enforcer and probe information is only generated when all the sub routines are active. 

So now, we keep trying to probe information from KubeArmor till we get info out or context times out.

We also have a slow animation while waiting so user know something is in process.
image

### DIFF
--- a/install/install.go
+++ b/install/install.go
@@ -105,7 +105,7 @@ func printAnimation(msg string, flag bool) int {
 	if flag {
 		progress++
 	}
-	printBar("    KubeArmor Installing ", 16)
+	printBar("    KubeArmor Installing ", 17)
 	return 0
 }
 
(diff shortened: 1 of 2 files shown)


install/install.go
```diff
@@ -141,23 +141,41 @@ func checkPods(c *k8s.Client, o Options) {
...
-		if k.ActiveLSM == "" || !k.ContainerSecurity {
-			enforcing = false
-			break
+	fmt.Print("\n🔧  Verifying KubeArmor functionality (this may take upto a minute) ...")
+	ctx, cancel := context.WithTimeout(context.Background(), 40*time.Second)
+	defer cancel()
+
+	for {
+		select {
+		case <-time.After(1 * time.Second):
```
daemon1024, line 150: It does not actually run after 1s, the process of probeRead in itself takes 5-6 second. I have seen anywhere between 2-4 attempts before we get the data. So set the total timeout to 40 seconds.

kranurag7: Barun, on a machine with a slower internet connection because I've seen even after 4 steps, the output is missing the infos? Should we increase this to 10sec or more? 
Also, it fixes https://github.com/kubearmor/kubearmor-client/issues/297

kranurag7: Yes, I tested the PR. I think the existing method was good too, in which no delay was there after all the services were running. By using this, we are making an assumption that the user will invoke `karmor probe` immediately after installing KubeArmor which may not be the case. Open to more thoughts on this.

daemon1024: > By using this, we are making an assumption that the user will invoke `karmor probe` immediately after installing KubeArmor which may not be the case. 

Not sure what you mean by making an assumption, we want users to know whether on installation KubeArmor enforcement works or not. Ref #334

kranurag7: It works good. I missed the context. Thanks for the ref.

daemon1024: Anurag, do you think we need to increase the time duration still?

kranurag7: No, I think this is more than enough. Initially I thought we were doing this so that we can get some time till the data gets written to `/tmp/KarmorProbe.cfg` which was my misunderstanding. I was thinking more in lines of #297

nyrahul (approved):

PR #340, opened 2023-06-27
feat: Add OCI registry support for KubeArmor policies

akshay196:
Part of https://github.com/kubearmor/KubeArmor/issues/1130

Support Karmor to interacting with OCI registry for pushing, pulling KubeArmor policies.

**Tasks**
- [ ] Adding complete Push support
   - [ ] Validate the policy before pushing
   - [ ] Authenticate with OCI registry if required
- [ ] Adding complete Pull support
  - [ ] Validate the policy after pulling
  - [ ] Authenticate with OCI registry if required


nyrahul:
It would be great to document this. Can you do a small writeup on:
1. What was the problem statement for doing this?
2. How would users benefit?
3. How would users use this?
4. Any refs etc.

PR #364, opened 2023-08-23
update the ginkgo tests

Ankurk99:
Update the tests to reflect actual number of policies after the removal of admission controller policy

:heavy_check_mark:  Tested locally on k3s.

### DIFF
--- a/tests/recommend/recommend_test.go
+++ b/tests/recommend/recommend_test.go
@@ -163,7 +163,7 @@ var _ = Describe("karmor", func() {
 				Expect(len(files)).To(BeNumerically(">=", 1))
 				Expect(err).To(BeNil())
 			})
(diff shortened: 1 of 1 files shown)


tests/recommend/recommend_test.go
```diff
@@ -163,7 +163,7 @@ var _ = Describe("karmor", func() {
 				Expect(len(files)).To(BeNumerically(">=", 1))
 				Expect(err).To(BeNil())
 			})
-			It("should contain atleast `5` policy files under directory `wordpress-mysql-wordpress` and should match with the files under `res/out/wordpress-mysql-wordpress`", func() {
+			It("should contain atleast `4` policy files under directory `wordpress-mysql-wordpress` and should match with the files under `res/out/wordpress-mysql-wordpress`", func() {
```
nyrahul, line 166: Hey, do you know the reason why now we have 4 policies in place of 5? The opposite would have been ok but reduction in policies should be understood.

Ankurk99:
Closing as the fix is not related to the actual issue.

PR #366, opened 2023-08-29
fixes: karmor probe

Ankurk99:
This PR added some fixes related to PR: https://github.com/kubearmor/kubearmor-client/pull/360

### DIFF
--- a/probe/probe.go
+++ b/probe/probe.go
@@ -46,7 +46,7 @@ var itwhite = color.New(color.Italic).Add(color.Italic).SprintFunc()
 // K8sInstaller for karmor install
 func probeDaemonInstaller(c *k8s.Client, o Options, krnhdr bool) error {
 	daemonset := deployment.GenerateDaemonSet(o.Namespace, krnhdr)
-	if _, err := c.K8sClientset.AppsV1().DaemonSets("").Create(context.Background(), daemonset, metav1.CreateOptions{}); err != nil {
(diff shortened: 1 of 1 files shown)


daemon1024 (changes requested):
This is unrelated. the probe daemonset installer and uninstaller are only triggered when KubeArmor is not present. 

But the probe issue is that it's not able to find kubearmor pod for extracting the information itself, even when KubeArmor is installed.

Ankurk99:
> This is unrelated. the probe daemonset installer and uninstaller are only triggered when KubeArmor is not present.

Yes, I am aware about this. Those changes were made in the previous PR and are reversed in this PR.

> But the probe issue is that it's not able to find kubearmor pod for extracting the information itself, even when KubeArmor is installed.

Please check the change at line 489. It addresses that concern.

probe/probe.go
```diff
@@ -486,7 +486,7 @@ func readDataFromKubeArmor(c *k8s.Client, o Options, nodeName string) (KubeArmor
 	cmdArr := []string{"cat", srcPath}
 	req := c.K8sClientset.CoreV1().RESTClient().
 		Get().
-		Namespace("").
+		Namespace(o.Namespace).
```
daemon1024, line 489:
```suggestion
		Namespace(pods.Items[0].Metadata.Namespace).
```

Please refer the PodSpec for exact path but this should solve it.

daemon1024 (changes requested):
My bad, regardless karmor probe shouldn't depend on namespace. It should be able to probe in any namespace. Below requested change should help with that.

daemon1024 (approved):

PR #368, opened 2023-09-01
(feat) Ability export profile data

PrimalPimmy:
New Feature added for profile! Now we can use the flag `karmor profile --save` to run profile, and as it runs, it sends the data to 4 separate json files for File, Process, Network and Summary respectively. All these files are stored in a common folder named `Profile_Summary`

### DIFF
--- a/profile/Client/profileClient.go
+++ b/profile/Client/profileClient.go
@@ -8,6 +8,10 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
+	"os"
+	"strings"
+	"time"
+
 	"github.com/accuknox/auto-policy-discovery/src/common"
 	"github.com/charmbracelet/bubbles/help"
 	"github.com/charmbracelet/bubbles/key"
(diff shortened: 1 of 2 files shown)


profile/Client/profileClient.go
```diff
@@ -425,8 +452,30 @@ func generateRowsFromData(data []pb.Log, Operation string) []table.Row {
...
+		if Operation == "File" {
+			convertToJson("File", jsondata)
+		} else if Operation == "Process" {
+			convertToJson("Process", jsondata)
+		} else if Operation == "Network" {
+			convertToJson("Network", jsondata)
+		} else if Operation == "Syscall" {
+			convertToJson("Syscall", jsondata)
+		}
+	}
```
daemon1024, line 477: How does this work? Will it only save when we are watching it?

PrimalPimmy: yup, if `--save` flag is there, then only it will watch and save it

daemon1024: Not that, does it save a single operation at a time?

PrimalPimmy: Nope, it saves all of them together.

PrimalPimmy: ref for generating table: https://github.com/kubearmor/kubearmor-client/blob/main/profile/Client/profileClient.go#L230

daemon1024: Thanks, Gotit.

PR #371, opened 2023-09-08
fix for downloading policy-templete zip

Prateeknandle:
removed `grab` package dependency for downloading policy-template zip file.
`grab.Get` gives error `server returned 403 Forbidden`

using `net/http` package instead.
for ref - https://accuknox.slack.com/archives/C0229TD83RA/p1694106383879899

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -45,7 +45,6 @@ require (
 require (
 	github.com/accuknox/auto-policy-discovery/src v0.0.0-20230707054448-845969c25277
 	github.com/accuknox/auto-policy-discovery/src/protobuf v0.0.0-20230707054448-845969c25277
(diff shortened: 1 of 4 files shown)


daemon1024 (changes requested):
Can we fallback to using local zip if somehow the new zip download/extraction fails?

Prateeknandle:
@daemon1024 where should we store the zip?

daemon1024:
@Prateeknandle I meant that

https://github.com/kubearmor/kubearmor-client/blob/11d6d3e1c30b15faf8a6790747540579e195c9ad/recommend/recommend.go#L108

in this line, let's not return error but just log a Warning and continue recommend with old templates

daemon1024 (approved):

PR #382, opened 2023-11-03
fix `karmor uninstall`

Ankurk99:
**Description**
- Uninstall KubeArmor Daemonset
- Uninstall all the resources based on label (kubearmor-app)
- `karmor uninstall` resources uninstall it from all the ns and
- `karmor uninstall -n <ns>` uninstalls the resources from that particular ns
- fixed the rotating status cursor issue on terminal resize

### DIFF
--- a/cmd/uninstall.go
+++ b/cmd/uninstall.go
@@ -24,7 +24,7 @@ var uninstallCmd = &cobra.Command{
 func init() {
 	rootCmd.AddCommand(uninstallCmd)
 
-	uninstallCmd.Flags().StringVarP(&uninstallOptions.Namespace, "namespace", "n", "kubearmor", "Namespace for resources")
+	uninstallCmd.Flags().StringVarP(&uninstallOptions.Namespace, "namespace", "n", "", "If no namespace is specified, it defaults to all namespaces and deletes all KubeArmor objects across them.")
 	uninstallCmd.Flags().BoolVar(&uninstallOptions.Force, "force", false, "Force remove KubeArmor annotations from deployments. (Deployments might be restarted)")
 	uninstallCmd.Flags().BoolVar(&uninstallOptions.Verify, "verify", true, "Verify whether all KubeArmor resources are cleaned up or not")
 }
--- a/install/install.go
+++ b/install/install.go
@@ -12,6 +12,7 @@ import (
 	"fmt"
 	"os"
 	"path"
+	"slices"
 	"strings"
 	"time"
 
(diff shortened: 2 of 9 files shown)


install/install.go
```diff
@@ -756,35 +793,59 @@ func K8sUninstaller(c *k8s.Client, o Options) error {
 	}
 
 	fmt.Print("❌   Cluster Role ...\n")
-	if err := c.K8sClientset.RbacV1().ClusterRoles().Delete(context.Background(), deployments.KubeArmorClusterRoleName, metav1.DeleteOptions{}); err != nil {
-		if !strings.Contains(err.Error(), "not found") {
-			return err
+	clusterRoleList, err := c.K8sClientset.RbacV1().ClusterRoles().List(context.TODO(), metav1.ListOptions{LabelSelector: "kubearmor-app"})
```
daemon1024, line 796: Is this backwards compatible?
We don't have the label selector in the current Cluster Role right?

daemon1024 (changes requested):

cmd/uninstall.go
```diff
@@ -24,7 +24,7 @@ var uninstallCmd = &cobra.Command{
 func init() {
 	rootCmd.AddCommand(uninstallCmd)
 
-	uninstallCmd.Flags().StringVarP(&uninstallOptions.Namespace, "namespace", "n", "kubearmor", "Namespace for resources")
+	uninstallCmd.Flags().StringVarP(&uninstallOptions.Namespace, "namespace", "n", "", "Namespace for resources")
```
DelusionalOptimist, line 27: I think we should add to the help message what would the default behavior be when no namespace is specified? :thinking:

install/install.go
```diff
@@ -633,181 +635,348 @@ func removeAnnotations(c *k8s.Client) {
...
 }
 
+func contains(slice []string, item string) bool {
+	for _, s := range slice {
+		if s == item {
+			return true
+		}
+	}
+	return false
+}
```
daemon1024, line 645: Use https://pkg.go.dev/slices#Contains

daemon1024 (changes requested):
small nit, tried it out locally. Works as intended. LGTM 🙌🏽

install/install.go
```diff
@@ -12,6 +12,7 @@ import (
 	"fmt"
 	"os"
 	"path"
+	"slices"
```
DelusionalOptimist, line 15: I think this package is still in experimental before Go 1.21.
Either need to update go.mod and CI workflows OR use `"golang.org/x/exp/slices"`

DelusionalOptimist (approved):
LGTM! Thanks @Ankurk99. :rocket:

daemon1024 (approved):

Issue #385, opened 2023-11-27
karmor probe panic


daemon1024:
https://github.com/kubearmor/kubearmor-client/blob/7b409be695906356db6f6f0f9beb42fd8245e350/probe/probe.go#L767

We have seen panics on this line of the code reported by users. Potentially the map access should be validated before appending the data.

![image](https://github.com/kubearmor/kubearmor-client/assets/47106543/822d91f9-1c27-44b4-b29e-150a96fed04e)

20RitikSingh:
I want to work on this.

akagami-harsh:
@daemon1024, i can see the PR is not yet merged can i give it a try

partik03:
@daemon1024 Please assign it to me

Manik2708:
Is this issue still alive?

daemon1024:
Nope. Thanks @partik03 for the fix!

PR #387, opened 2023-11-27
fix the karmor probe panic

MeenuyD:
Safely access the map in pod processing

Updated the code to safely access the map 'mp' in pod processing to prevent potential panics when retrieving 'NsPostureString' and 'NsVisibilityString' values. Added a check to verify the existence of the key before accessing it.
#385

### DIFF
--- a/probe/probe.go
+++ b/probe/probe.go
@@ -764,7 +764,12 @@ func getAnnotatedPods(c *k8s.Client, o Options, postureData map[string]string) (
 			if err != nil {
 				return nil, [][]string{}, err
 			}
-			data = append(data, []string{armoredPod.Namespace, mp[armoredPod.Namespace].NsPostureString, mp[armoredPod.Namespace].NsVisibilityString, armoredPod.Name, ""})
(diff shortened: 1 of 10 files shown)


probe/probe.go
```diff
@@ -764,7 +764,12 @@ func getAnnotatedPods(c *k8s.Client, o Options, postureData map[string]string) (
 			if err != nil {
 				return nil, [][]string{}, err
 			}
-			data = append(data, []string{armoredPod.Namespace, mp[armoredPod.Namespace].NsPostureString, mp[armoredPod.Namespace].NsVisibilityString, armoredPod.Name, ""})
+			nsPostureString, nsVisibilityString := "", ""
+			if nsData, exists := mp[armoredPod.Namespace]; exists {
```
daemon1024, line 768:
```suggestion
			if nsData, exists := mp[p.Namespace]; exists {
```
This might be out of scope of the issue, but can we check why are we getting details of the armoredPod seperately instead of using the pod object passed down in the for loop?

Issue #386, opened 2023-11-27
`karmor install --env=generic --save` should work without kubernetes cluster access


daemon1024:
With the updates in `karmor install` It tries to create namespace if not found.

This logic prevents user without kubernetes to just generate the manifest. If `--save` is used, karmor should not fail if k8s env is not found, it should just go ahead and use the provided env or use the generic env to generate the manifest file.

![image](https://github.com/kubearmor/kubearmor-client/assets/47106543/3cee1c5e-c2fe-4379-a1b7-7a56e12adebe)

sheharyaar:
I would like to work on this one.

sheharyaar:
created a PR, please review.

daemon1024:
Thanks @sheharyaar again for handling it. Appreciate it.

PR #388, opened 2023-11-28
ignore cluster access with save flag

sheharyaar:


sheharyaar:
fixes #386

install/install.go
```diff
@@ -236,7 +241,12 @@ func K8sInstaller(c *k8s.Client, o Options) error {
 			},
 		}
 		if _, err := c.K8sClientset.CoreV1().Namespaces().Create(context.Background(), &newns, metav1.CreateOptions{}); err != nil {
-			return fmt.Errorf("failed to create namespace %s: %+v", ns, err)
+			if o.Save {
+				fmt.Printf("⚠️\tfailed to create namespace %s: %+v\n", ns, err)
```
nyrahul, line 245: I think if `save` is enabled then karmor should not be creating any k8s resources including namespace ... 

The logic should be to bypass namespace create if `save` is attempted.

@daemon1024 WDYT?

nyrahul (changes requested):

install/install.go
daemon1024: That's correct.

sheharyaar: Ok, will do the required changes

daemon1024 (approved):
LGTM, Thanks for the quick handling.

Issue #376, opened 2023-10-04
Update Long description for karmor install


ShubhamTatvamasi:
At present, the `karmor install` command shares an identical description with both the Shot and Long versions. We require a more comprehensive and detailed description that thoroughly explains the command's functionality.

https://github.com/kubearmor/kubearmor-client/blob/main/cmd/install.go#L18:L19

nyrahul:
Not only `install` but other options can also be targeted.

ShubhamTatvamasi:
> I would like to give this a shot, could you please assign this to me?

Sure, Done.

S3DFX-CYBER:
@ShubhamTatvamasi is this issue fixed or is it open for assignment?

PR #392, opened 2023-12-11
Improve the long description for install,profile,log and policy commands

harisudarsan1:
This PR solves the issue #376  by adding a long description for install command.


rksharma95:
@harisudarsan1 can you please update the PR, and let's target other commands as well if it needs to update description.

rootxrishabh:
@harisudarsan1, please rebase

PR #396, opened 2024-01-04
Shift installation to helm

rootxrishabh:
Purpose:
This PR shifts `karmor install` and `karmor uninstall` to use kubearmor operator using helm.

Does this PR introduce a breaking change?
Yes

Checklist:

- [x] Configure `verify` flag
- [x] Implement better logs visuals for install and uninstall.
- [x] Fix installation throwing error when namespace `kubearmor` already exists.


rootxrishabh:
Folks, still working on this to implement/fix functions.

rootxrishabh:
Squashed commits for DCO, namespace errors have been resolved.

install/install.go
```diff
@@ -210,385 +218,93 @@ func checkTerminatingPods(c *k8s.Client, ns string) int {
 	return 0
 }
 
-// K8sInstaller for karmor install
-func K8sInstaller(c *k8s.Client, o Options) error {
```
daemon1024, line 214: Let's not remove it entirely, let's keep it under the legacy flag. It's still convenient for quick testing where operator is not needed.

daemon1024 (changes requested):
Thanks a lot for the PR. Had one comment on a quick overview.

daemon1024:
Fixes https://github.com/kubearmor/KubeArmor/issues/1256

install/install.go
rootxrishabh: Sounds good! I'll create a flag for legacy install and uninstall.

rootxrishabh: @daemon1024 If a user installs using `--legacy true` then karmor uninstall will throw an error as legacy installation doesn't create helm release. To enforce legacy uninstall I see 2 options - 

1. Inform the user to try legacy uninstall using` karmor uninstall --legacy=true` when helm throws error.
2. When using legacy installation. Create a label on the daemonset `legacy=true` which will be used to automatically uninstall using the legacy approach.
**The second approach does complicate it a bit but automates the uninstallation while the first one is self-explanatory.
Eager to know your thoughts!**

daemon1024: Let's go with first method. It's fine to do that.

rootxrishabh:
Folks, as of I have only implemented the image and init-image flag. Instead of using `autoDeploy=true `we are now using operator clientset to create the CR with values. Also, we cannot do multiple helm revisions when using the operator client set.
Looking for reviews on the flag implementation approach : )

rootxrishabh:
One more thing, the [sample-config](https://github.com/kubearmor/KubeArmor/blob/main/pkg/KubeArmorOperator/config/samples/sample-config.yml) does not state the [KubeRbacProxyImage](https://github.com/kubearmor/KubeArmor/blob/main/pkg/KubeArmorOperator/api/operator.kubearmor.com/v1/kubearmorconfig_types.go#L45), should we include it in the flags?

rootxrishabh:
Waiting on #[1589](https://github.com/kubearmor/KubeArmor/issues/1589) and #[1590](https://github.com/kubearmor/KubeArmor/issues/1590) to be addressed before flags are implemented.

PR #400, opened 2024-02-01
Enable custom registry.

achrefbensaad:
**Purpose:**

Enable kubearmor installation from a custom registry, this is useful in cases where the user wants to use a private registry rather than ours.

**How to test:**

push the following images to a custom registry:
- kubearmor/kubearmor-relay-server:latest
- kubearmor/kubearmor:stable
- kubearmor/kubearmor-init:stable
- gcr.io/kubebuilder/kube-rbac-proxy:v0.8.0
- kubearmor/kubearmor-controller:latest

use `-r` flag while installing.


daemon1024 (changes requested):
how would the operator handles this?
// @rksharma95 @rootxrishabh

rksharma95:
> how would the operator handles this?

There's no default registry set with the operator and it's possible to provide the image reference along with custom registry using KubearmorConfig CR.

daemon1024 (approved):

Issue #398, opened 2024-01-21
Local Operator installation doesn’t change posture settings when supplied in kubearmorconfig


rootxrishabh:
## Bug Report

**General Information**

- Environment description - K3s
- Kernel version - `Linux kubearmor-os-1 6.2.0-1019-gcp kubearmor/KubeArmor#21~22.04.1-Ubuntu SMP Thu Nov 16 18:18:34 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux`
- Orchestration system version in use - v1.23
- Link to relevant artifacts (policies, deployments scripts, ...)
- Target containers/pods

**To Reproduce**

1. Set default posture settings to block.
![image](https://github.com/kubearmor/KubeArmor/assets/73490663/ea039736-5779-4210-8eda-828550c77cb0)

2. Deploy Kubearmor using helm-based local deployment.
![image](https://github.com/kubearmor/KubeArmor/assets/73490663/d1efab95-acf2-4974-815a-c507dd05f8ee)

3. Confirm posture settings using `karmor probe`.
![image](https://github.com/kubearmor/KubeArmor/assets/73490663/1e334415-e83f-4526-9d49-970d2d9547f1)

**Expected behavior**
Karmor probe should confirm that Default Posture is set to block based for File, Capabilities, and Network but rather shows audit.
CC @rksharma95

PR #402, opened 2024-02-08
Enabled operator installation for karmor

rootxrishabh:
Purpose:
This PR shifts karmor install and karmor uninstall to use kubearmor operator using helm.

Does this PR introduce a breaking change?
Yes


Aryan-sharma11:
Snyk issues remediation:-
- Upgrade github.com/opencontainers/runc/libcontainer to version 1.1.12 or higher. ( and all pkgs related to it)
- Upgrade github.com/lestrrat-go/jwx/v2/jws to version 2.0.19 or higher.
- Upgrade go.etcd.io/etcd/v3/server/lease to version 3.4.29, 3.5.11 or higher
- Upgrade go.etcd.io/etcd/v3/server/etcdserver to version 3.4.29, 3.5.11 or higher.

There are still some pkgs whose fixes are pushed to the main branch but not published yet.
- github.com/urfave/negroni/v2 introduced though github.com/sigstore/timestamp-authority@v1.2.0

daemon1024:
Fixes https://github.com/kubearmor/KubeArmor/issues/1256

cmd/uninstall.go
```diff
@@ -16,15 +16,24 @@ var uninstallCmd = &cobra.Command{
...
-		return err
+		if uninstallOptions.Legacy {
+			if err := install.K8sLegacyUninstaller(client, uninstallOptions); err != nil {
+				return err
+			}
+		} else {
+			if err := install.K8sUninstaller(client, uninstallOptions); err != nil {
+				return err
+			}
+		}
```
daemon1024, line 27: We need to auto fallback to legacy uninstallation, if we don't find helm release.

Most of the uninstallations today are going to be through legacy mode, it's going to be a bad UX/DX if we ask them to use legacy flag here

cmd/uninstall.go
```diff
@@ -16,15 +16,24 @@ var uninstallCmd = &cobra.Command{
...
+		return nil
 	},
 }
 
 func init() {
 	rootCmd.AddCommand(uninstallCmd)
 
-	uninstallCmd.Flags().StringVarP(&uninstallOptions.Namespace, "namespace", "n", "", "If no namespace is specified, it defaults to all namespaces and deletes all KubeArmor objects across them.")
+	uninstallCmd.Flags().StringVarP(&uninstallOptions.Namespace, "namespace", "n", "kubearmor", "If no namespace is specified, it defaults to all namespaces and deletes all KubeArmor objects across them.")
```
daemon1024, line 35: uninstall is meant to be namespace agnostic. It will be responsibility of the client to find which namespace the installation is in provided namespace is not provided.

install/install.go
```diff
@@ -128,23 +191,41 @@ func checkPods(c *k8s.Client, o Options) {
...
-			fmt.Printf("\r⌚️\tCheck Incomplete due to Time-Out!                     \n")
-			break
+
+	if !o.Legacy {
+		for {
+			time.Sleep(900 * time.Millisecond)
+			pods, _ := c.K8sClientset.CoreV1().Pods("").List(context.TODO(), metav1.ListOptions{LabelSelector: "kubearmor-app=kubearmor", FieldSelector: "status.phase==Running"})
+			podno := len(pods.Items)
+			fmt.Printf("\rℹ️\tWaiting for KubeArmor Daemonset to run:  %s", cursor[cursorcount])
+			cursorcount++
```
daemon1024, line 201: Let's have independent checkpods function for legacy installation. Let's not couple both the logic in if else.

daemon1024 (changes requested):
Overall code LGTM. Great Work.

I have some enhancement suggestions inline.

Let's fix the conflicts as well.

Can you include screenshots/asciinema video of how the installation and uninstallation process looks like?

install/install.go
```diff
@@ -636,6 +714,159 @@ func K8sInstaller(c *k8s.Client, o Options) error {
...
+	var printYAML []interface{}
+
+	kubearmorConfig := getOperatorCR(o)
+
+	if o.Save {
+		printYAML = append(printYAML, kubearmorConfig)
+
+		currDir, err := os.Getwd()
+		if err != nil {
+			return err
```
daemon1024, line 744: o.Save is meant to help with a dry run of what the CLI is going to install.

It's meant to be same as https://helm.sh/docs/helm/helm_template/

Just saving the CR is not enough, we need to generate the entire manifest with the operator YAML as well.

daemon1024 (changes requested):

cmd/uninstall.go
rootxrishabh: Done.

cmd/uninstall.go
rootxrishabh: Done :+1:

daemon1024 (changes requested):

daemon1024:
`karmor install --save` failing if no kubernetes cluster is configured
We do not mandate K8s cluster if we only need to save

image

daemon1024 (commented):
I don't think the first log is looking nice, is it something we need to keep?

image

daemon1024:
Registry flag is not working

```
./karmor install -r "ttl.sh" --save
```

Expected - config images to be prepended with registry info
Actual - No change in config

daemon1024:
```
./karmor install --tag=v1.2.1 --save
```

Not working as expected

Expected - config images and operator image to have their tags changed
Actual - No change in config

rootxrishabh:
> I couldn't help but notice, even if I am just using save. It takes a lot of time to generate things. Is it a drawback of using helm client?

I tried printing manifests without running the client and it returns a null pointer. So, the install/upgrade client had to run dryrun in order to get the manifests.
Taking a look at other points :eyes:

rootxrishabh:
> `karmor install --save` failing if no kubernetes cluster is configured We do not mandate K8s cluster if we only need to save
> 
> image

`rootxrishabh@rootxrishabh:~/kubearmor-client$ ./karmor install --legacy=true
😄      Environment : k3s                                                                 
🔥      CRD kubearmorpolicies.security.kubearmor.com                                      
ℹ️       CRD kubearmorpolicies.security.kubearmor.com already exists                       
🔥      CRD kubearmorhostpolicies.security.kubearmor.com                                  
ℹ️       CRD kubearmorhostpolicies.security.kubearmor.com already exists                   
💫      Service Account                                                                   
💫      Service Account                                                                   
⚙️       Cluster Role                                                                      
ℹ️       Cluster Role already exists                                                       
⚙️       Cluster Role Bindings                                                             
ℹ️       Cluster Role Bindings already exists                                              
⚙️       KubeArmor Relay Roles                                                             
🛡       KubeArmor Relay Service                                                           
🛰       KubeArmor Relay Deployment                                                        
🛡       KubeArmor DaemonSet - Init kubearmor/kubearmor-init:stable, Container kubearmor/kubearmor:stable  -gRPC=32767  
ℹ️       KubeArmor DaemonSet already exists                                                
🛡       KubeArmor Controller TLS certificates                                             
💫      KubeArmor Controller Service Account                                              
⚙️       KubeArmor Controller Roles                                                        
🚀      KubeArmor Controller Deployment                                                   
🚀      KubeArmor Controller Metrics Service                                              
🚀      KubeArmor Controller Webhook Service                                              
🥳      Done Installing KubeArmor                                                         
🤩      KubeArmor Controller Mutation Admission Registration                              
ℹ️       KubeArmor Controller Mutation Admission Registration already exists               88%
🚀      KubeArmor ConfigMap Creation                                                      88%
😋      Checking if KubeArmor pods are running...▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇] 111.76%
🥳      Done Checking , ALL Services are running!             
⌚️      Execution Time : 13.858858021s 

🔧      Verifying KubeArmor functionality (this may take upto a minute)...

        🛡️       Your Cluster is Armored Up! 
rootxrishabh@rootxrishabh:~/kubearmor-client$ ./karmor uninstall
❌      KubeArmor is either not installed, or the specified namespace is incorrect.
ℹ️       Please ensure you have installed KubeArmor, and check that you are specifying the correct namespace.
ℹ️       Attempting legacy uninstallation.
🗑️  Mutation Admission Registration
🗑️  KubeArmor Services
    ❌  Service: kubearmor removed
    ❌  Service: kubearmor-controller-metrics-service removed
💨  Service Accounts
    ❌  ServiceAccount kubearmor removed
    ❌  ServiceAccount kubearmor-relay removed
    ❌  ServiceAccount kubearmor-controller removed
💨  Cluster Roles
    ❌  ClusterRole kubearmor-snitch removed
    ❌  ClusterRole kubearmor-clusterrole removed
    ❌  ClusterRole kubearmor-relay-clusterrole removed
    ❌  ClusterRole kubearmor-controller-proxy-role removed
    ❌  ClusterRole kubearmor-controller-clusterrole removed
💨  Cluster Role Bindings
    ❌  ClusterRoleBinding kubearmor-clusterrolebinding removed
    ❌  ClusterRoleBinding kubearmor-relay-clusterrolebinding removed
    ❌  ClusterRoleBinding kubearmor-controller-clusterrolebinding removed
    ❌  ClusterRoleBinding kubearmor-controller-proxy-rolebinding removed
    ❌  ClusterRoleBinding kubearmor-snitch-binding removed
🧹  Roles
    ❌  Role kubearmor-controller-leader-election-role removed
🧹  RoleBindings
    ❌  RoleBinding kubearmor-controller-leader-election-rolebinding removed
👻  KubeArmor Controller TLS certificates
    ❌  KubeArmor Controller TLS certificate kubearmor-controller-webhook-server-cert removed
👻  KubeArmor ConfigMap
    ❌  ConfigMap kubearmor-config removed
👻  KubeArmor DaemonSet
    ❌  KubeArmor DaemonSet kubearmor removed
👻  KubeArmor Deployments
    ❌  KubeArmor Deployment kubearmor-relay removed
    ❌  KubeArmor Deployment kubearmor-controller removed
🔄  Checking if KubeArmor pods are stopped...
🔴  Done Checking; all services are stopped!             
⌚️  Termination Time: 9.787902842s `
Yes

daemon1024:
image

Installation Process in general is smooooth. Great Work.

But this does not look cohesive. 
- the helm logs and emoji logs are not streamlines, We should suppress helm logs and manually log based on the output maybe. Not sure what's best here.
- We are spamming `Deployment is not ready` messages
- We need to type something when KubeArmor Daemonset is deploying, Because it takes a while, including messaging around expected time, inform that snitch is running

cmd/uninstall.go
```diff
@@ -27,4 +31,5 @@ func init() {
 	uninstallCmd.Flags().StringVarP(&uninstallOptions.Namespace, "namespace", "n", "", "If no namespace is specified, it defaults to all namespaces and deletes all KubeArmor objects across them.")
 	uninstallCmd.Flags().BoolVar(&uninstallOptions.Force, "force", false, "Force remove KubeArmor annotations from deployments. (Deployments might be restarted)")
 	uninstallCmd.Flags().BoolVar(&uninstallOptions.Verify, "verify", true, "Verify whether all KubeArmor resources are cleaned up or not")
+	uninstallCmd.Flags().BoolVar(&uninstallOptions.Legacy, "legacy", false, "Legacy uninstallation for kubearmor")
```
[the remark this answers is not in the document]
rootxrishabh: Oops, this one was accidently left out when I implemented auto fallback to legacy uninstallation, will remove.

daemon1024:
```
karmor uninstall
```

This is not working as expected
I did the standard helm install
But it is saying it cannot detect it
image
It fellback to legacy uninstallation, but my helm package still exists

rootxrishabh:
> image
> 
> Installation Process in general is smooooth. Great Work.
> 
> But this does not look cohesive.
> 
>     * the helm logs and emoji logs are not streamlines, We should suppress helm logs and manually log based on the output maybe. Not sure what's best here.
> 
>     * We are spamming `Deployment is not ready` messages
> 
>     * We need to type something when KubeArmor Daemonset is deploying, Because it takes a while, including messaging around expected time, inform that snitch is running

I think we should handle logs ourselves and suppress helm logs. We could manually output logs for deployments etc.
I will add deployment message for snitch and include an expected time logs.

rootxrishabh:
> ```
> karmor uninstall
> ```
> 
> This is not working as expected I did the standard helm install But it is saying it cannot detect it image It fellback to legacy uninstallation, but my helm package still exists

Yes, I was thinking of discussing this earlier. If we remove the default namespace in uninstall and don't specify namespace for example "kubearmor" then our current helm implementation doesn't automatically detect the installed namespace. I guess implementing auto detection of namespace is required?

cmd/uninstall.go
```diff
@@ -16,8 +16,12 @@ var uninstallCmd = &cobra.Command{
 	Short: "Uninstall KubeArmor from a Kubernetes Cluster",
 	Long:  `Uninstall KubeArmor from a Kubernetes Clusters`,
 	RunE: func(cmd *cobra.Command, args []string) error {
-		err := install.K8sUninstaller(client, uninstallOptions)
-		return err
+		if err := install.K8sUninstaller(client, uninstallOptions); err != nil {
+			if err := install.K8sLegacyUninstaller(client, uninstallOptions); err != nil {
+				return err
+			}
+		}
```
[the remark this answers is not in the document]
rootxrishabh: Yes, screenshot attached.

daemon1024:
> I guess implementing auto detection of namespace is required?

Yup. We can list packages and go through that ig.

rootxrishabh:
> I don't think the first log is looking nice, is it something we need to keep?
> 
> image

Should we suppress logs for save flag?

daemon1024:
I think it will automatically be handled when we have the entire logging figured out for helm.

rootxrishabh:
> ```
> ./karmor install --tag=v1.2.1 --save
> ```
> 
> Not working as expected
> 
> Expected - config images and operator image to have their tags changed Actual - No change in config

We currently dont have --save watching tag and registry flags. I will do it :+1:

daemon1024:
> We currently dont have --save watching tag and registry flags. I will do it 👍

That's concerning. Ideal case it should be not be 2 different approaches. We finalise on the config then decide whether to save it or install it.

rootxrishabh:
> > We currently dont have --save watching tag and registry flags. I will do it 👍
> 
> That's concerning. Ideal case it should be not be 2 different approaches. We finalise on the config then decide whether to save it or install it.

Got it. I will make sure we are handling all flags first.

rootxrishabh:
> I couldn't help but notice, even if I am just using save. It takes a lot of time to generate things. Is it a drawback of using helm client?

Update: I fixed --save not running for clusterless environments and now the process is running client side only, executes without helm logs and is quick.

rootxrishabh:
Here's a demo of the functionality - 
kubearmorgif

daemon1024:
We should be updating the operator image as well when we use tag/flags/registry
image

operator image is not changing with any flags. We might need to impement a new flag to update operator image. I believe the helm chart accepts updates to operator image as well

rootxrishabh:
We can now change imagePullPolicy, tag, image, registry.
image

cmd/install.go
DelusionalOptimist, line 1: For users, like devs, who want to have advanced configuration in the CR, doing it all through flags will make the command difficult to use, how about a `--skip-deploy` flag with default value false? If someone specifies it, `KubeArmorConfig` CR won't be created and maybe just given on stdout or saved to a file with some instructions to use it, so that they can modify it according to their use and deploy manually.

cmd/install.go
```diff
@@ -43,7 +47,6 @@ func init() {
 	installCmd.Flags().BoolVar(&installOptions.Save, "save", false, "Save KubeArmor Manifest ")
 	installCmd.Flags().BoolVar(&installOptions.Verify, "verify", true, "Verify whether all KubeArmor resources are created, running and also probes whether KubeArmor has armored the cluster or not")
```
DelusionalOptimist, line 48: Good to have: If needed, let's mark these flags as [mutually exclusive](https://pkg.go.dev/github.com/spf13/cobra#Command.MarkFlagsMutuallyExclusive) to fail fast.

install/install.go
```diff
@@ -49,35 +61,101 @@ type Options struct {
...
-	Environment string
-}
+func getOperatorCR(o Options) *Operatorv1.KubeArmorConfig {
+	ns := o.Namespace
+	if o.Tag != "" {
+		o.KubearmorImage = updateImageTag(o.KubearmorImage, o.Tag)
+		o.InitImage = updateImageTag(o.InitImage, o.Tag)
+		o.ControllerImage = updateImageTag(o.ControllerImage, o.Tag)
+		o.RelayImage = updateImageTag(o.RelayImage, o.Tag)
+	}
```
DelusionalOptimist, line 74: Someone might want to specify a different tag for only one of the images but same for all other.
For example, if I run
```
$ ./karmor install --tag v1.0 --relay-image=kubearmor/kubearmor-relay-server:stable --save
```
the generated KubeArmorConfig CR should have `v1.0` tag for all the images but `kubearmor/kubearmor-relay-server:stable` for `kubearmorRelayImage`.
If no tag is given in the specified image, for e.g. someone specified only `kubearmor/kubearmor-relay-server` take it from `--tag`.

install/install.go
```diff
@@ -49,35 +61,101 @@ type Options struct {
...
+	if o.Block == "all" || strings.Contains(o.Block, "file") {
+		defaultFilePosture = "block"
+	}
+	if o.Block == "all" || strings.Contains(o.Block, "network") {
+		defaultNetworkPosture = "block"
+	}
+	if o.Block == "all" || strings.Contains(o.Block, "capabilities") {
+		defaultCapabilitiesPosture = "block"
+	}
```
DelusionalOptimist, line 109: Let's refactor this if possible.

install/install.go
```diff
@@ -660,11 +821,205 @@ func K8sInstaller(c *k8s.Client, o Options) error {
...
+	// Install the CR using operator clientset
+
+	operatorClientSet, err := operatorClient.NewForConfig(c.Config)
+	if err != nil {
+		return fmt.Errorf("failed to create operator client: %w", err)
+	}
+	if _, err := operatorClientSet.OperatorV1().KubeArmorConfigs(ns).Get(context.Background(), "kubearmorconfig-default", metav1.GetOptions{}); err == nil {
+		fmt.Println("ℹ️\tKubeArmorConfig already exists")
+	} else {
+		if _, err := operatorClientSet.OperatorV1().KubeArmorConfigs(ns).Create(context.Background(), kubearmorConfig, metav1.CreateOptions{}); err != nil {
```
DelusionalOptimist, line 1010: Maybe we can use [IsAlreadyExists](https://pkg.go.dev/k8s.io/apimachinery/pkg/api/errors#IsAlreadyExists) here.

install/install.go
```diff
@@ -660,11 +821,205 @@ func K8sInstaller(c *k8s.Client, o Options) error {
...
+			return fmt.Errorf("failed to create KubeArmorConfig CR: %w", err)
+		}
+		fmt.Println("😄\tKubeArmorConfig Created")
+	}
+
+	if o.Verify && !o.Save {
+		checkPods(c, o, client.Install)
+	}
+
+	return err
```
DelusionalOptimist, line 1020: Not a good practice to return a nil value variable. Let's return nil.
```suggestion
	return nil
```

install/install.go
```diff
@@ -1051,6 +1405,68 @@ func K8sUninstaller(c *k8s.Client, o Options) error {
 	return nil
 }
 
+// K8sUninstaller for karmor uninstall
+func K8sUninstaller(c *k8s.Client, o Options) error {
```
DelusionalOptimist, line 1409: Since the `kubearmor-operator` helm release is installed first, then KubeArmor. Any reason for listing pods, instead of present helm releases for checking if KubeArmor is installed or not?

It might lead to situations like below:
Operator is in `imagePullBackOff` and no other kubearmor resources exist, unable to uninstall.
image

install/install.go
```diff
@@ -1051,6 +1405,68 @@ func K8sUninstaller(c *k8s.Client, o Options) error {
...
+	}
+
+	settings := cli.New()
+	settings.SetNamespace(ns)
+	settings.RESTClientGetter()
+
+	actionConfig := new(action.Configuration)
+	if err := actionConfig.Init(settings.RESTClientGetter(), ns, os.Getenv("HELM_DRIVER"), log.Printf); err != nil {
+		log.Printf("%+v", err)
+		os.Exit(1)
```
DelusionalOptimist, line 1437: Why not return error?

install/install.go
```diff
@@ -1051,6 +1405,68 @@ func K8sUninstaller(c *k8s.Client, o Options) error {
...
+	if err != nil {
+		fmt.Println("ℹ️  Helm release not found. Switching to legacy uninstaller.")
+		return err
+	}
+
+	operatorClientSet, err := operatorClient.NewForConfig(c.Config)
+	if err != nil {
+		return fmt.Errorf("failed to create operator client: %w", err)
+	}
+	if err := operatorClientSet.OperatorV1().KubeArmorConfigs(ns).Delete(context.Background(), "kubearmorconfig-default", metav1.DeleteOptions{}); err != nil {
```
DelusionalOptimist, line 1457: We might not want to return an error if the resource itself was not found. Let's use [IsNotFound](https://pkg.go.dev/k8s.io/apimachinery/pkg/api/errors#IsNotFound) so that rest of the process can continue.

install/install.go
```diff
@@ -1051,6 +1405,68 @@ func K8sUninstaller(c *k8s.Client, o Options) error {
...
+		return fmt.Errorf("failed to delete KubeArmorConfig CR: %w", err)
+	} else {
+		fmt.Println("❌\tAll KubeArmor resources removed")
+	}
+
```
DelusionalOptimist, line 1462: Can we have the force flag for removing annotations and CR/CRDs? Ref - https://github.com/kubearmor/kubearmor-client/blob/main/install/install.go

install/install.go
```diff
@@ -660,11 +821,205 @@ func K8sInstaller(c *k8s.Client, o Options) error {
...
+	r.CachePath = settings.RepositoryCache
+
+	if _, err := r.DownloadIndexFile(); err != nil {
+		return fmt.Errorf("failed to download repository index file: %w", err)
+	}
+
+	var repoFile repo.File
+	repoFile.Update(entry)
+	if err := repoFile.WriteFile(settings.RepositoryConfig, 0644); err != nil {
```
DelusionalOptimist, line 935: Question: Do we know where this file gets written to? Does it get cleaned up later?

install/install.go
```diff
@@ -660,11 +821,205 @@ func K8sInstaller(c *k8s.Client, o Options) error {
...
+	chartRequested, err := loader.Load(chartPath)
+	if err != nil {
+		return fmt.Errorf("failed to load Helm chart: %w", err)
+	}
+	log.SetOutput(io.Discard)
+	upgradeInstaller, err := client.Run(releaseName, chartRequested, values)
+	log.SetOutput(os.Stdout)
+	if err != nil {
+		client.Install = true
+		if client.Install {
```
DelusionalOptimist, line 961: Is the check needed?

DelusionalOptimist (changes requested):
Awesome work @rootxrishabh! It's coming along just how we thought! :rocket: 
Some suggestions/questions inline. PTAL.

DelusionalOptimist:
Also the uninstall command stays stuck right now because of - https://github.com/kubearmor/KubeArmor/issues/1638

rootxrishabh:
@DelusionalOptimist can we address the review changes first and then take a look at the controller issue?

install/install.go
rootxrishabh: `check.Install` is purely an informative flag. However, we pass it to the checkPods function to check whether the snitch has to be inspected or not. The snitch can only be checked at install and not at Helm upgrade, as the snitch job is meant to be erased after installation, leaving no way to track it during upgrades.

cmd/install.go
rootxrishabh: Sounds good!
Suggestion: Currently, `karmor install --save` does not output values for [posture](https://github.com/kubearmor/kubearmor-client/blob/main/cmd/install.go#L40) and [visibility](https://github.com/kubearmor/kubearmor-client/blob/main/cmd/install.go#L42) due to the absence of default values for audit and visibility in karmor. I think we should mention them anyway because the operator adds the same when a default value is found to be null. Also, mentioning default values won't affect legacy installations. WDYT?

rootxrishabh: Also, I think we should create a separate file for this and not use `kubearmor.yaml`.

install/install.go
rootxrishabh: Instead of the print statement?

install/install.go
rootxrishabh: Oops, Already have a function to handle this, will be addressed in the commit. 👍

install/install.go
rootxrishabh: This code piece basically performs `helm repo add kubearmor https://kubearmor.github.io/charts` and `helm repo update kubearmor`. It is stored in `$HOME/.config/helm/repositories.yaml` and it does not clean up after uninstall. Happens same with `helm uninstall`.

install/install.go
rootxrishabh: Great catch! Thank you. I will refactor the checking mechanism.

install/install.go
rootxrishabh: Yes, we can perform these operations using clientset/operatorclientset. I'll implement these.

install/install.go
rootxrishabh: Update: The [if](https://github.com/kubearmor/kubearmor-client/blob/e5913634d9e286d36f536be256169547bb953290/install/install.go#L1412) block is first checking for user supplied namespace correctness and the [else](https://github.com/kubearmor/kubearmor-client/blob/e5913634d9e286d36f536be256169547bb953290/install/install.go#L1419) block is auto-finding the namespace kubearmor is installed in if the user did not specify a namespace. Once namespace is determined [it](https://github.com/kubearmor/kubearmor-client/blob/e5913634d9e286d36f536be256169547bb953290/install/install.go#L1445) checks weather the installation was done through helm and procceds to choose standard or legacy uninstallation.

install/install.go
rootxrishabh: Done 👍

cmd/install.go
rootxrishabh: I have configured it to output to stdout, didn't want kubeaarmor.yaml to clash with this and making a new file seemed too much for this.

PrimalPimmy:
@rootxrishabh the defaultposture flags are not working for me 
image

image

I used block capabilities as flag but it is still in audit mode

rootxrishabh:
@PrimalPimmy can you please check the kubearmorconfig-default configmap. Currently posture and visibility settings are being incorrectly displayed by karmor probe for operator installation.
[Ref](https://github.com/kubearmor/kubearmor-client/issues/398)

PrimalPimmy:
> @PrimalPimmy can you please check the kubearmor-config configmap. Currently posture and visibility settings are being incorrectly displayed by karmor probe for operator installation.

Still not working 
image

file is not going to audit mode when set?

rootxrishabh:
> I see that the kubearmorconfig is not being recreated. You need to do `karmor uninstall --force` to remove the CR and then reinstall with the posture settings.

install/install.go
```diff
@@ -39,46 +50,107 @@ type Options struct {
...
+			Namespace: ns,
+			Labels: map[string]string{
+				"app.kubernetes.io/name":       "kubearmorconfig",
+				"app.kubernetes.io/instance":   "kubearmorconfig-default",
+				"app.kubernetes.io/part-of":    "kubearmoroperator",
+				"app.kubernetes.io/managed-by": "kustomize",
+				"app.kubernetes.io/created-by": "kubearmoroperator",
+			},
+		},
+		Spec: Operatorv1.KubeArmorConfigSpec{
```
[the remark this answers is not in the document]
rootxrishabh: I think it is doable. I'll take a look!

Aryan-sharma11:
Snyk issues remediation:-
- Upgrade helm.sh/helm/v3/pkg to version 3.14.2 or higher.
- Upgrade github.com/opencontainers/runc/libcontainer to version 1.1.12 or higher, introduced through
`github.com/Microsoft/hcsshim@v0.11.4`.
- Upgrade github.com/lestrrat-go/jwx/v2/jws to version 2.0.19 or higher, Introduced through
`github.com/Microsoft/hcsshim@v0.11.4` .
- Upgrade go.etcd.io/etcd/v3/server to version 3.4.29, 3.5.11 or higher, Introduced through
`github.com/google/certificate-transparency-go@v1.1.7`.
- for `github.com/urfave/negroni` introduced through `github.com/sigstore/timestamp-authority@v1.2.0 › github.com/urfave/negroni@v1.0.0` fix pushed to master branch but not yet published .
no fixed version :-
- for helm.sh/helm/v3/cmd/helm.

PrimalPimmy:
> I see that the kubearmorconfig is not being recreated. You need to do karmor uninstall --force to remove the CR and then reinstall with the posture settings.

Is there any way to get the configmap recreated, imo it's a better way. Maybe remove the configmap after uninstalling

rootxrishabh:
> > I see that the kubearmorconfig is not being recreated. You need to do karmor uninstall --force to remove the CR and then reinstall with the posture settings.
> 
> Is there any way to get the configmap recreated, imo it's a better way. Maybe remove the configmap after uninstalling

The configmap kubearmor-config does get deleted but the operator CR kubearmorconfig-default does not and so for every new installation after this, the CR does not get updated and thus we don't see posture settings supplied through CLI. Inorder to remove the CR(kubearmorconfig-default), we must do `karmor uninstall --force` and do a fresh installation.
[Ref](https://github.com/kubearmor/kubearmor-client/pull/402#discussion_r1493333511)

rootxrishabh:
> > > I see that the kubearmorconfig is not being recreated. You need to do karmor uninstall --force to remove the CR and then reinstall with the posture settings.
> > 
> > 
> > Is there any way to get the configmap recreated, imo it's a better way. Maybe remove the configmap after uninstalling
> 
> The configmap kubearmor-config does get deleted but the operator CR kubearmorconfig-default does not and so for every new installation after this, the CR does not get updated and thus we don't see posture settings supplied through CLI. Inorder to remove the CR(kubearmorconfig-default), we must do `karmor uninstall --force` and do a fresh installation. [Ref](https://github.com/kubearmor/kubearmor-client/pull/402#discussion_r1493333511)

We can notice it outputs KubeArmorConfig created in a fresh installation. 
image

daemon1024:
If I do a `kubectl apply -f` with the update config, it does update existing config if there are changes. I believe the same behaviour should translate here

image

daemon1024:
We should switch to doing
```go
err:= Resource.Create()
if !strings.Contains(err.Error(), "already exists") {
		err:= Resource.Create()
		return err/nil
}
```

rootxrishabh:
> If I do a `kubectl apply -f` with the update config, it does update existing config if there are changes. I believe the same behaviour should translate here
> 
> image

I remember us discussing the idea of `karmor update`. I think we discussed that we would update all operator CR options separately using the update command, it can be handled in a separate issue. Wdyt?

daemon1024:
> I remember us discussing the idea of `karmor update`. I think we discussed that we would update all operator CR options separately using the update command, it can be handled in a separate issue. Wdyt?


This is a different issue. People won't realise that CR is not updated. the update command was meant to patch things as well patch the helm release if needed.

We cannot ask people to run uninstall with force just so that they have a fresh installation.

It's a 2 liner change here so I believe it's worth it to handle here. Wdyt?

rootxrishabh:
> > I remember us discussing the idea of `karmor update`. I think we discussed that we would update all operator CR options separately using the update command, it can be handled in a separate issue. Wdyt?
> 
> This is a different issue. People won't realise that CR is not updated. the update command was meant to patch things as well patch the helm release if needed.
> 
> We cannot ask people to run uninstall with force just so that they have a fresh installation.
> 
> It's a 2 liner change here so I believe it's worth it to handle here. Wdyt?

Agreed. `--force` installation shouldn't be an intermediatory to update the CR. I will implement the CR updation here.

install/install.go
```diff
@@ -126,11 +209,97 @@ func printMessage(msg string, flag bool) int {
...
+	otime := stime.Add(600 * time.Second)
+	cursor := [4]string{"|", "/", "—", "\\"}
+	// Check snitch completion only for install
+	if i {
+		for {
+			time.Sleep(900 * time.Millisecond)
+			pods, _ := c.K8sClientset.CoreV1().Pods("").List(context.TODO(), metav1.ListOptions{LabelSelector: "kubearmor-app=kubearmor-snitch", FieldSelector: "status.phase==Succeeded"})
+			podno := len(pods.Items)
+			remaining := time.Until(otime).Round(time.Second)
+			fmt.Printf("\rℹ️\tWaiting for KubeArmor Snitch to run:  %s, estimated time remaining: %v", cursor[cursorcount], remaining)
```
daemon1024, line 223: I don't it's a good idea to display this esitmated time. 
Let's just mention, this may take a couple of minutes.
Also this is not cleared up after it actually succeeds

image

install/install.go
```diff
@@ -1063,6 +1434,62 @@ func K8sUninstaller(c *k8s.Client, o Options) error {
...
+
+		fmt.Printf("❌\tRemoving CRD %s\n", kocName)
+		if err := c.APIextClientset.ApiextensionsV1().CustomResourceDefinitions().Delete(context.Background(), kocName, metav1.DeleteOptions{}); apierrors.IsNotFound(err) {
+			fmt.Printf("CRD %s not found\n", kocName)
+		}
+
+		removeAnnotations(c, ns)
+	}
+
+	fmt.Println("❌\tKubeArmor resources removed")
```
daemon1024, line 1484: Can we add a statement here, 
```
If !o.Force
Resources not managed by helm/GLobal Resources are not cleaned up. Please use karmor uninstall --force if you want complete cleanup
```

WDYT?

daemon1024 (changes requested):

install/install.go
rootxrishabh: Sounds good. I will take a look.

install/install.go
rootxrishabh: @daemon1024 does this look good?
Currently not able to figure out the exact of reason of  `\` residue
Screenshot from 2024-03-12 11-32-33

rootxrishabh: Screenshot from 2024-03-12 18-17-49

install/install.go
```diff
@@ -265,19 +432,16 @@ func K8sInstaller(c *k8s.Client, o Options) error {
...
-			} else {
-				return errors.New("unsupported environment or cluster not configured correctly")
-			}
+	env = k8s.AutoDetectEnvironment(c)
+	if env == "none" {
+		if o.Save {
+			printMessage("⚠️\tNo env provided with \"--save\", setting env to  \"generic\"", true)
+			env = "generic"
+		} else {
+			return errors.New("unsupported environment or cluster not configured correctly")
```
daemon1024, line 441: Why was this done?
We could still want to save a legacy yaml without wanting to autodetect it.

install/install.go
```diff
@@ -1063,6 +1432,64 @@ func K8sUninstaller(c *k8s.Client, o Options) error {
...
+	log.SetOutput(io.Discard)
+	_, err = client.Run("kubearmor-operator")
+	log.SetOutput(os.Stdout)
+	if err != nil {
+		fmt.Println("ℹ️   Error uninstalling through Helm. Switching to legacy uninstaller.")
+		return err
+	}
+
+	if !o.Force {
+		fmt.Println("ℹ️   Resources not managed by helm/GLobal Resources are not cleaned up. Please use karmor uninstall --force if you want complete cleanup.")
```
daemon1024, line 1464:
```suggestion
		fmt.Println("ℹ️   Resources not managed by helm/Global Resources are not cleaned up. Please use karmor uninstall --force if you want complete cleanup.")
```

daemon1024 (changes requested):
LGTM. One question inline and one nit.

install/install.go
rootxrishabh: We have removed the env flag in this PR. We can have it enabled again if we want to give the user an option to generate kubearmor.yaml using the env they might provide.

rootxrishabh: I have implemented manual env passing for --save.

install/install.go
```diff
@@ -265,19 +446,23 @@ func K8sInstaller(c *k8s.Client, o Options) error {
...
-	if o.Env.Auto {
-		env = k8s.AutoDetectEnvironment(c)
-		if env == "none" {
-			if o.Save {
-				printMessage("⚠️\tNo env provided with \"--save\", setting env to  \"generic\"", true)
-				env = "generic"
-			} else {
-				return errors.New("unsupported environment or cluster not configured correctly")
-			}
+	env = k8s.AutoDetectEnvironment(c)
```
daemon1024, line 449: Can't we just roll back to the way it was earlier? I see a lot of edge case issues here like save won't work when k8s client is missing

cmd/install.go
```diff
@@ -34,19 +37,24 @@ func init() {
...
 	installCmd.Flags().BoolVar(&installOptions.Local, "local", false, "Use Local KubeArmor Images (sets ImagePullPolicy to 'IfNotPresent') ")
-	installCmd.Flags().StringVarP(&installOptions.Env.Environment, "env", "e", "", "Supported KubeArmor Environment [k0s,k3s,microK8s,minikube,gke,bottlerocket,eks,docker,oke,generic]")
 	installCmd.Flags().StringVarP(&installOptions.ImageRegistry, "registry", "r", "", "Image registry to use to pull the images")
+	installCmd.Flags().BoolVar(&installOptions.Legacy, "legacy", false, "Installs kubearmor in legacy mode if set to true")
+	installCmd.Flags().BoolVar(&installOptions.SkipDeploy, "skip-deploy", false, "Saves kubearmor operator CR manifest rather than deploying it")
 	installCmd.Flags().BoolVar(&installOptions.PreserveUpstream, "preserve-upstream", true, "Do not override the image registry when using -r flag, prefix only")
+	installCmd.Flags().StringVarP(&installOptions.Env, "env", "e", "", "Supported KubeArmor Environment [k0s,k3s,microK8s,minikube,gke,bottlerocket,eks,docker,oke,generic], only works with --save")
```
daemon1024, line 57: We do not need it to work with only save. Let's not change any behaviour with legacy installation.

cmd/install.go
```diff
@@ -34,19 +37,24 @@ func init() {
...
 	installCmd.Flags().StringVarP(&installOptions.Block, "block", "b", "", "Kubearmor Block Posture Context [all,file,network,capabilities]")
 	installCmd.Flags().StringVarP(&installOptions.Visibility, "viz", "", "", "Kubearmor Telemetry Visibility [process,file,network,none]")
 	installCmd.Flags().BoolVar(&installOptions.Save, "save", false, "Save KubeArmor Manifest ")
 	installCmd.Flags().BoolVar(&installOptions.Verify, "verify", true, "Verify whether all KubeArmor resources are created, running and also probes whether KubeArmor has armored the cluster or not")
 	installCmd.Flags().BoolVar(&installOptions.Local, "local", false, "Use Local KubeArmor Images (sets ImagePullPolicy to 'IfNotPresent') ")
-	installCmd.Flags().StringVarP(&installOptions.Env.Environment, "env", "e", "", "Supported KubeArmor Environment [k0s,k3s,microK8s,minikube,gke,bottlerocket,eks,docker,oke,generic]")
 	installCmd.Flags().StringVarP(&installOptions.ImageRegistry, "registry", "r", "", "Image registry to use to pull the images")
+	installCmd.Flags().BoolVar(&installOptions.Legacy, "legacy", false, "Installs kubearmor in legacy mode if set to true")
```
daemon1024, line 54: Let's mark legacy and env as deprecated flags. You can check cobra docs to see how to mark something as deprecated.

daemon1024 (changes requested):

go.mod
```diff
@@ -1,72 +1,70 @@
 module github.com/kubearmor/kubearmor-client
 
-go 1.21
+go 1.21.0
```
daemon1024, line 3:
```suggestion
go 1.21
```
Let's not ping go version

daemon1024 (approved):

PR #408, opened 2024-03-05
feat(gRPC): add mtls support

rksharma95:
This PR adds mtls support to secure grpc connections between `karmor <-> KubeArmor` and `karmor <-> KubeArmor-Relay`.

The following configurations has been added:
|config|values|default| description |
|---------|---------|-----------|------------------|
|`--insecure`|true/false|true| connect to secure/insecure server |
|`--tlsCertPath`|"valid directory path"|/var/lib/kubearmor/tls| path to the directory where ca.crt, client.crt and client.key files are present |
|`--tlsCertProvider`|self/external|self| self: client certificates will be generated dynamically, external: use client and ca certificates provided using files |

### Use cases:

1. KubeArmor is being deployed in k8s cluster with tls enabled, connect to the kubearmor relay to get the telemetry using default self signed ca deployed along with the KubeArmor. (**KubeArmor should be installed with tls configurations that allows karmor to connect**)

```
karmor logs --insecure=false
```
2. Connect to KubeArmor using certificates `(client.crt, client.key, ca.crt)` stored locally.

```
karmor logs --insecure=false --tlsCertPath=<path-to-certs> --tlsCertProvider=external
```


log/logClient.go
```diff
@@ -100,17 +103,28 @@ type Feeder struct {
...
+	if !o.Insecure {
+		tlsCreds, err := loadTLSCredentials(c, o)
+		if err != nil {
+			return nil
+		}
+		creds = tlsCreds
+	} else {
+		creds = insecure.NewCredentials()
+	}
+	conn, err := grpc.Dial(fd.server, grpc.WithTransportCredentials(creds))
```
daemon1024, line 125: Can we try to fallback to secure if grpc.Dial fails because server rejected insecure connection?

log/tls.go
```diff
@@ -0,0 +1,70 @@
...
+	"k8s.io/klog/v2"
+)
+
+var (
+	KubeArmorCALabels = map[string]string{
+		"kubearmor-app": "kubearmor-ca",
+	}
+)
+
+func getKubeArmorCaSecret(client kubernetes.Interface) (string, string) {
```
daemon1024, line 26: let's keep these functions in the k8s package not in log

daemon1024 (changes requested):
Please handle gosec failure.

log/logClient.go
rksharma95: if server has tls enabled the fallback logic will not work.

log/log.go
```diff
@@ -144,9 +153,18 @@ func StartObserver(c *k8s.Client, o Options) error {
 	}
 
 	// create client
-	logClient := NewClient(gRPC, o.MsgPath, o.LogPath, o.LogFilter, o.Limit)
-	if logClient == nil {
-		return errors.New("unable to create log client")
+	logClient, err := NewClient(gRPC, o, c.K8sClientset)
+	if err != nil {
+		if o.Insecure {
```
daemon1024, line 158: We should ignore doing the retry if the error is due to invalid address

log/log.go
```diff
@@ -144,9 +153,18 @@ func StartObserver(c *k8s.Client, o Options) error {
...
+			// retry connecting to the server on secured channel
+			o.Insecure = false
+			logClient, _ = NewClient(gRPC, o, c.K8sClientset)
```
daemon1024, line 161: Let's log the earlier error maybe

daemon1024 (changes requested):

log/log.go
rksharma95: we're doing so right!

daemon1024 (approved):
LGTM. Will merge after #396

PR #416, opened 2024-03-17
Remove usage of sudo to install kubearmor-client

### LINKED ISSUE #415: Install karmor without `sudo`
### Description
Currently to install kubearmor-client one has to run the below which fetches and runs https://github.com/kubearmor/kubearmor-client/blob/main/install.sh and installs karmor at the specified path.
```
curl -sfL http://get.kubearmor.io/ | sudo sh -s -- -b <path-to-install>
```

We set the path to `/usr/local/bin` everywhere in our [docs](https://github.com/kubearmor/KubeArmor/blob/main/getting-started/deployment_guide.md#install-karmor-cli-optional) so that user's can run it easily. However, for this we need to use sudo while executing the script which is not desirable from a security POV.

### Proposal
- Instead of using `/usr/local/bin`, use a path writable by user for e.g `$HOME/.local/bin/karmor` everywhere in our docs and remove `sudo`.
- Make the install script emit a command so that users can add the above path to their `PATH` variable either in the current shell or in their shell's RC.

lekaf974:
Fix #415

Result
```bash
[lekaf974@~ ]$ curl -sfL http://get.kubearmor.io/ | sh -s -- -b "$HOME/.local/bin"
kubearmor/kubearmor-client info checking GitHub for latest tag
kubearmor/kubearmor-client info found version: 1.2.0 for v1.2.0/linux/amd64
kubearmor/kubearmor-client info installed /home/lekaf974/.local/bin/karmor
kubearmor/kubearmor-client info Add /home/lekaf974/.local/bin to your PATH
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@
 ## Installation
 
 ```shell
-curl -sfL http://get.kubearmor.io/ | sudo sh -s -- -b /usr/local/bin
+`curl -sfL http://get.kubearmor.io/ | sh -s` 
 ```
 
 ### Installing From Source
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@
 ## Installation
 
 ```shell
-`curl -sfL http://get.kubearmor.io/ | sh -s` 
+curl -sfL http://get.kubearmor.io/ | sh -s
 ```
 
 ### Installing From Source
--- a/install.sh
+++ b/install.sh
@@ -60,6 +60,8 @@ execute() {
     fi
     install "${srcdir}/${binexe}" "${BINDIR}/"
     log_info "installed ${BINDIR}/${binexe}"
+    log_info "karmor is installed in ${BINDIR}"
+    log_info "invoke ${BINDIR}/${binexe} or move karmor to your desired PATH"
   done
   rm -rf "${tmpdir}"
 }


kranurag7 (commented):
This way we are asking users to add `~/.local/bin` which a user might not want. 

Can we use install it in bin directory of the current working directory? 
And print a log message to move `karmor` to any location which is already present in PATH. 

`mv ./bin/karmor ~/.local/bin` or any other path as per user preference. 

If we intend to use `~/.local/bin` then the PR looks good to me.

lekaf974:
> This way we are asking users to add `~/.local/bin` which a user might not want.
> 
> Can we use install it in bin directory of the current working directory? And print a log message to move `karmor` to any location which is already present in PATH.
> 
> `mv ./bin/karmor ~/.local/bin` or any other path as per user preference.
> 
> If we intend to use `~/.local/bin` then the PR looks good to me.

Just needed to validate, I replaced `/usr/local/bin` to `~/.local.bin` just to be able to install without sudo.
I took `~/.local/bin` since `-b` option was already present. 
I can simply remove the option that way the binary will be installed in `./bin` which is the default behavior in the script

If it makes sense let me know so I'll make proper change.

@DelusionalOptimist any preference ?

install.sh
```diff
@@ -60,6 +60,7 @@ execute() {
     fi
     install "${srcdir}/${binexe}" "${BINDIR}/"
     log_info "installed ${BINDIR}/${binexe}"
+    log_info "Add ${BINDIR} to PATH variable"
```
kranurag7, line 63: @lekaf974 Sorry for getting back a little late on this one. I think instead of we asking to update the PATH on the user system, we can change the message and let them decide where they want to move. 

So, I think you change of installing karmor in `./bin/karmor` looks good to me and I will change the message to something like `karmor is installed in ./bin/karmor` and `invoke ./bin/karmor or move karmor to your desired PATH` 

It's an open discussion and we can discuss and agree on what works best. 
Let me know if you've any follow-up questions here.

lekaf974: LGTM I don't have a real opinion, I'll update the message so the user can decide what he want to do soon!

lekaf974: Latest output
```bash
kubearmor/kubearmor-client info checking GitHub for latest tag
kubearmor/kubearmor-client info found version: 1.2.1 for v1.2.1/linux/amd64
kubearmor/kubearmor-client info installed ./bin/karmor
kubearmor/kubearmor-client info karmor is installed in ./bin
kubearmor/kubearmor-client info invoke ./bin/karmor or move karmor to your desired PATH
```

lekaf974: @kranurag7 ready for another review

kranurag7: From my side, this looks good to me. Thank you for working on this.

README.md
```diff
@@ -8,7 +8,7 @@
 ## Installation
 
 ```shell
-curl -sfL http://get.kubearmor.io/ | sudo sh -s -- -b /usr/local/bin
+`curl -sfL http://get.kubearmor.io/ | sh -s`
```
daemon1024, line 11:
```suggestion
curl -sfL http://get.kubearmor.io/ | sh -s
```

It's already inside a code blocks :D

daemon1024 (changes requested):
Looks Good. Thanks a lot. Sorry this has taken so long.

One minor nit, We can merge this after that :D

PR #438, opened 2024-05-21
fix(probe): directly fetch kubearmor pods from k8s api instead of listing nodes

### LINKED ISSUE #435: Improve `karmor probe` error handling
### Description
Probe panics on any error. Certain scenarios:
- if it doesn't find KubeArmor on one of the nodes
- If KubeArmor pods are in pending state
etc.

A recent panic:
image

### Remediation
Improve error handling in `karmor probe`

daemon1024:
this approach fixes panic in case kubearmor is not running on a node for some reason

Fixes #435

### DIFF
--- a/probe/probe.go
+++ b/probe/probe.go
@@ -422,7 +422,8 @@ func getKubeArmorDaemonset(c *k8s.Client, o Options) (bool, *Status) {
 		return false, nil
 	}
 	desired, ready, available := w.Items[0].Status.DesiredNumberScheduled, w.Items[0].Status.NumberReady, w.Items[0].Status.NumberAvailable
-	if desired != ready && desired != available {
+	if desired != ready && desired != available && ready == 0 {
+		// set kubearmor to not running only if there are 0 ready pods
 		return false, nil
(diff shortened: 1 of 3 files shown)


probe/probe.go
```diff
@@ -487,19 +487,21 @@ func getKubeArmorContainers(c *k8s.Client, o Options) map[string]*KubeArmorPodSp
...
 	var dataList []KubeArmorProbeData
-	for i, item := range nodes.Items {
-		data, err := readDataFromKubeArmor(c, o, item.Name)
+	for i, item := range pods.Items {
+		if item.Status.Phase != corev1.PodRunning {
+			continue
+		}
+		data, err := readDataFromKubeArmor(c, item)
 		if err != nil {
 			return []KubeArmorProbeData{}, nil, err
```
DelusionalOptimist, line 506: An error for any one of the nodes will do an early return here. We won't get probe output of other nodes. Is it possible to handle this?
image

DelusionalOptimist (approved):
LGTM

Issue #439, opened 2024-06-05
JSON output from `karmor logs --json` is not beautified


navin772:
Currently `karmor logs --json` gives a cluttered output in a single line:
```json
{"Timestamp":1717591294,"UpdatedTime":"2024-06-05T12:41:34.673689Z","ClusterName":"default","HostName":"navin","NamespaceName":"wordpress-mysql","Owner":{"Ref":"Deployment","Name":"wordpress","Namespace":"wordpress-mysql"},"PodName":"wordpress-78bc585459-626nf","Labels":"app=wordpress","ContainerID":"26f3f479957b89e219806d62df04234a8f1660029346d11239fa9fd86f7a3cc9","ContainerName":"wordpress","ContainerImage":"wordpress:4.8-apache@sha256:6216f64ab88fc51d311e38c7f69ca3f9aaba621492b4f1fa93ddf63093768845","HostPPID":39028,"HostPID":39354,"PPID":193,"PID":199,"UID":0,"ProcessName":"/usr/bin/curl","PolicyName":"DefaultPosture","Type":"MatchedPolicy","Source":"/usr/bin/curl goo.gl","Operation":"Network","Resource":"protocol=UDP,type=SOCK_DGRAM","Data":"lsm=SOCKET_CREATE protocol=UDP,type=SOCK_DGRAM","Enforcer":"BPFLSM","Action":"Block","Result":"Permission denied","Cwd":"/"}
```
By using `MarshalIndent` instead of `Marshal` we can get beautified output like this:

```json
{
  "Timestamp": 1717591662,
  "UpdatedTime": "2024-06-05T12:47:42.500041Z",
  "ClusterName": "default",
  "HostName": "navin",
  "NamespaceName": "wordpress-mysql",
  "Owner": {
    "Ref": "Deployment",
    "Name": "wordpress",
    "Namespace": "wordpress-mysql"
  },
  "PodName": "wordpress-78bc585459-626nf",
  "Labels": "app=wordpress",
  "ContainerID": "26f3f479957b89e219806d62df04234a8f1660029346d11239fa9fd86f7a3cc9",
  "ContainerName": "wordpress",
  "ContainerImage": "wordpress:4.8-apache@sha256:6216f64ab88fc51d311e38c7f69ca3f9aaba621492b4f1fa93ddf63093768845",
  "HostPPID": 39028,
  "HostPID": 43125,
  "PPID": 193,
  "PID": 201,
  "UID": 0,
[...]

PR #446, opened 2024-06-25
introduced an output flag

EraKin575:
introduced option flags to  beautify logs

fixes: https://github.com/kubearmor/kubearmor-client/issues/439


navin772:
Hi @EraKin575 thanks for the PR but I cannot seem to get the desired output with `karmor logs --output pretty-json`, it just prints the normal output as in `karmor logs`.

You need to change the [`WatchTelemetryHelper`](https://github.com/kubearmor/kubearmor-client/blob/51dda84f3f29d81c7ff5a82a5a35a62fc0ec1ba4/log/logClient.go#L310) function.
Also, for the `-o` shorthand use `StringVarP()` instead of `StringVar()`. 

In this process, I ended up solving the issue. I can guide you with the solution or if you don't mind I can open a PR and you can give your feedback there.

EraKin575:
@navin772 Sorry, I got stuck in some stuff. I have fixed the output flag. thanks for the assist

EraKin575:
closing due to duplicate PRs

PR #449, opened 2024-07-23
probe added to sysdump

### LINKED ISSUE #305: Add `karmor probe` as a part of `karmor sysdump`
**Description**
Currently `karmor sysdump` generated output doesn't contains information about the node's support for KubeArmor (available LSMs, the mode of enforcement, etc.) which are already a part of `karmor probe`. 
The aim is to get that output as a part of `karmor sysdump`

Manik2708:
Fixes #305 
Probe output is now being saved to `karmor.probe` in zip file.

### DIFF
--- a/cmd/probe.go
+++ b/cmd/probe.go
@@ -34,6 +34,6 @@ func init() {
 	probeCmd.Flags().StringVarP(&probeInstallOptions.Namespace, "namespace", "n", "kubearmor", "Namespace for resources")
 	probeCmd.Flags().BoolVar(&probeInstallOptions.Full, "full", false, `If KubeArmor is not running, it deploys a daemonset to have access to more
 information on KubeArmor support in the environment and deletes daemonset after probing`)
-	probeCmd.Flags().StringVarP(&probeInstallOptions.Output, "format", "f", "text", " Format: json or text ")
+	probeCmd.Flags().StringVarP(&probeInstallOptions.Output, "format", "f", "text", "Format: json or text or no-color")
 	probeCmd.Flags().StringVar(&probeInstallOptions.GRPC, "gRPC", "", "GRPC port ")
 }
--- a/probe/print.go
+++ b/probe/print.go
@@ -1,6 +1,9 @@
 package probe
 
 import (
+	"fmt"
+	"io"
+	"log"
 	"os"
 	"strconv"
 
@@ -8,8 +11,8 @@ import (
 	"github.com/olekukonko/tablewriter"
 )
 
-func renderOutputInTableWithNoBorders(data [][]string) {
-	table := tablewriter.NewWriter(os.Stdout)
+func (o *Options) renderOutputInTableWithNoBorders(data [][]string) {
+	table := tablewriter.NewWriter(o.getWriter())
 	table.SetAutoWrapText(false)
 	table.SetAutoFormatHeaders(true)
 	table.SetHeaderAlignment(tablewriter.ALIGN_LEFT)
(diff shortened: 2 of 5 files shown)


Manik2708:
See this for reference!

Screenshot from 2024-07-23 20-15-19

nyrahul:
Thanks @Manik2708, Please handle the go-sec findings failing in the CI.

sysdump/sysdump.go
```diff
@@ -180,6 +186,34 @@ func Collect(c *k8s.Client, o Options) error {
...
+			GRPC:      o.GRPC,
+		})
+		if err != nil {
+			return err
+		}
+		writer.Close()
+		os.Stdout = oldStdOut
+		out, _ := io.ReadAll(reader)
+		// This is necessary to beautify the terminal output
+		ansiEscapePattern := `\x1b\[[0-9;]*m`
```
nyrahul, line 210: Rather than doing this, there might be value in adding a `karmor probe --nocolor` option?

Now I see the reason to change the `color` -> to escape sequence .. so that here the escape sequences can be removed. Is this right?

cmd/sysdump.go
```diff
@@ -26,4 +26,9 @@ var sysdumpCmd = &cobra.Command{
 func init() {
 	rootCmd.AddCommand(sysdumpCmd)
 	sysdumpCmd.Flags().StringVarP(&dumpOptions.Filename, "file", "f", "", "output file to use")
+	sysdumpCmd.Flags().StringVarP(&dumpOptions.Namespace, "namespace", "n", "kubearmor", "Namespace for resources")
```
nyrahul, line 29: I think there should be no reason to add this `namespace` option. `karmor probe` autodetects the namespace afaik.

nyrahul (changes requested):

probe/print.go
```diff
@@ -28,9 +29,8 @@ func renderOutputInTableWithNoBorders(data [][]string) {
 // printDaemonsetData function
 func printDaemonsetData(daemonsetStatus *Status) {
 	var data [][]string
-
-	color.Green("\nFound KubeArmor running in Kubernetes\n\n")
-	_, err := boldWhite.Printf("Daemonset :\n")
+	fmt.Fprintf(os.Stdout, "\033[32m\nFound KubeArmor running in Kubernetes\033[0m\n\n")
+	_, err := printWhiteBoldText("Daemonset :\n")
```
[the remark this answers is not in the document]
Manik2708: Yes! There is a reason! `color.Green()` is not able to save the output in `os.Stdout`, I moved into it's internal working and got to know that it is using `printf` which doesn't write in `os.Stdout` stream. See this!

Screenshot from 2024-07-23 22-52-54

But to save date in file we require it in stream!

sysdump/sysdump.go
Manik2708: Yes, we require it to remove some absurd characters from file!

cmd/sysdump.go
Manik2708: Ok gonna remove it!

Manik2708:
Sorry for the formatting error! Please re-run the workflow

sysdump/sysdump.go
Manik2708: So is the need is to use this pattern in a flag `nocolor` and then using this flag here?

Prateeknandle: @Manik2708  with `nocolor` flag we do not add any color or bold text, we print normal text and thus there will be no requirement for handling ANSI escape codes. In sysdump we set the [nocolor here](https://github.com/kubearmor/kubearmor-client/pull/449/files#diff-e9c40a8e7876ce0237b03aba3f2b4f81ad70f07ae0b8ff1cffa57e76ef5eec8fR196) to `true`.
does that sounds good?
(No styling is needed through sysdump)

probe/print.go
```diff
@@ -28,9 +29,8 @@ func renderOutputInTableWithNoBorders(data [][]string) {
...
 	if err != nil {
 		color.Red(" Error while printing")
```
Prateeknandle, line 35: will this cause issue when error hits, since we are using `color` mod?

Manik2708: No, it's just that the error will not be saved in file!

sysdump/sysdump.go
Manik2708: So, how will we remove all colors? Is there any package? Or we have to refactor?

Prateeknandle: I don't think there is any official package for this, any other package will probably do the same what we have in the pr, refactor is the only option I see here.

Manik2708: Ok, I will start working on refactoring that! Thanks!

Prateeknandle: let's have a print func and pass the print statement and `option.noColor` to it.

probe/print.go
Manik2708: I think it will come under the refactor! I will try to refactor all the `color` with a print function!

sysdump/sysdump.go
Aryan-sharma11: Instead of a separate `noColor`  option should we add a new output type?  let's say `o.output == "sysdump"`. wdyt @Prateeknandle @nyrahul @DelusionalOptimist

Prateeknandle: yes, @Aryan-sharma11 that'll be good to have, so output can take values `text, nocolor-text, json`. @Manik2708 please consider this.

Manik2708: That means `output` flag will take these values and give output accordingly! Am I right?

Prateeknandle: yes

Manik2708:
Please see the screenshots after the new commit:

Screenshot from 2024-07-27 00-02-10

Screenshot from 2024-07-27 00-01-59

probe/print.go
```diff
@@ -25,26 +26,46 @@ func renderOutputInTableWithNoBorders(data [][]string) {
...
+		_, err := fmt.Fprint(os.Stdout, s)
+		if err != nil {
+			_, printErr := red.Printf("Can't print to os.Stdout")
+			if printErr != nil {
+				fmt.Printf("Printing error")
+			}
+		}
+	} else {
+		_, err := c.Fprintf(os.Stdout, s)
+		if err != nil {
```
Aryan-sharma11, line 48: let's print the error too

Aryan-sharma11 (changes requested):

Aryan-sharma11:
@Manik2708, I tested these changes locally, there are some improvements we can make 
1) The karmor probe data being stored in file name `karmor.probe` instead let's make it `karmor-probe.txt`
2) in first line we can see `getting logs from kubearmor-apparmor-containerd-98c2c-q5jbb` we don't need to print this
image
3) this is the karmor probe output with your changes 
image
let's have the headings in bold text, for reference this is the current karmor probe's output
image

Manik2708:
> @Manik2708, I tested these changes locally, there are some improvements we can make
> 
>     1. The karmor probe data is being stored in file name `karmor.probe` instead let's make it `karmor-probe.txt`
> 
>     2. in first line we can see `getting logs from kubearmor-apparmor-containerd-98c2c-q5jbb` we don't need to print this
>        image
> 
>     3. this is the karmor probe output with your changes
>        image
>        let's have the headings in bold text, for reference this is the current karmor probe's output
>        image

@Aryan-sharma11 Have resolved all your suggestions. Please review! Thanks.

probe/probe.go
```diff
@@ -384,32 +382,30 @@ func probeNode(c *k8s.Client, o Options) {
 	nodes, _ := c.K8sClientset.CoreV1().Nodes().List(context.Background(), metav1.ListOptions{})
 	if len(nodes.Items) > 0 {
 		for i, item := range nodes.Items {
-			_, err := boldWhite.Printf("Node %d : \n", i+1)
-			if err != nil {
-				color.Red(" Error")
-			}
+			str := fmt.Sprintf("Node %d : \n", i+1)
+			o.printToOutput(itwhite, str)
```
Aryan-sharma11, line 386:
```suggestion
			o.printToOutput(boldwhite, str)
```

probe/probe.go
```diff
@@ -102,26 +105,24 @@ func PrintProbeResult(c *k8s.Client, o Options) error {
...
 
-			color.Green("\nFound KubeArmor running in Systemd mode \n\n")
+			o.printToOutput(green, "\nFound KubeArmor running in Systemd mode \n\n")
 
-			_, err := boldWhite.Printf("Host : \n")
-			if err != nil {
-				color.Red(" Error")
-			}
-			printKubeArmorProbeOutput(kd)
+			o.printToOutput(itwhite, "Host : \n")
```
Aryan-sharma11, line 110:
```suggestion
			o.printToOutput(boldwhite, "Host : \n")
```

Aryan-sharma11 (changes requested):

Manik2708:
Sorry for being late @Aryan-sharma11, I think output should be alright now!

cmd/probe.go
```diff
@@ -34,6 +34,6 @@ func init() {
 	probeCmd.Flags().StringVarP(&probeInstallOptions.Namespace, "namespace", "n", "kubearmor", "Namespace for resources")
 	probeCmd.Flags().BoolVar(&probeInstallOptions.Full, "full", false, `If KubeArmor is not running, it deploys a daemonset to have access to more
 information on KubeArmor support in the environment and deletes daemonset after probing`)
-	probeCmd.Flags().StringVarP(&probeInstallOptions.Output, "format", "f", "text", " Format: json or text ")
+	probeCmd.Flags().StringVarP(&probeInstallOptions.Output, "format", "f", "text", " Format: json or text or no-color")
```
Aryan-sharma11, line 37:
```suggestion
	probeCmd.Flags().StringVarP(&probeInstallOptions.Output, "format", "f", "text", "Format: json or text or no-color")
```

cmd/sysdump.go
```diff
@@ -26,4 +26,8 @@ var sysdumpCmd = &cobra.Command{
 func init() {
 	rootCmd.AddCommand(sysdumpCmd)
 	sysdumpCmd.Flags().StringVarP(&dumpOptions.Filename, "file", "f", "", "output file to use")
+	sysdumpCmd.Flags().BoolVar(&dumpOptions.Full, "full", false, `If KubeArmor is not running, it deploys a daemonset to have access to more
```
Aryan-sharma11, line 29: we should not add flags for probe here, instead pass the default ones.

cmd/sysdump.go
```diff
@@ -26,4 +26,8 @@ var sysdumpCmd = &cobra.Command{
...
+information on KubeArmor support in the environment and deletes daemonset after probing`)
+	sysdumpCmd.Flags().StringVarP(&dumpOptions.Output, "type", "t", "text", " Output type: json or text ")
```
Aryan-sharma11, line 31: flag not needed, let the default be no-color.

cmd/sysdump.go
```diff
@@ -26,4 +26,8 @@ var sysdumpCmd = &cobra.Command{
...
+	sysdumpCmd.Flags().StringVar(&dumpOptions.GRPC, "gRPC", "", "GRPC port ")
```
Aryan-sharma11, line 32: This is also not required as currently `sysdump` does not generate info if  kubearmor is running in systemd mode.

Aryan-sharma11 (changes requested):

Manik2708:
@Aryan-sharma11 @nyrahul @Prateeknandle @rootxrishabh Please review!

sysdump/sysdump.go
```diff
@@ -180,6 +181,43 @@ func Collect(c *k8s.Client, o Options) error {
...
+			return err
+		}
+		err = writer.Close()
+		if err != nil {
+			return err
+		}
+		os.Stdout = oldStdOut
+		out, _ := io.ReadAll(reader)
+		outStr := string(out)
+		outArr := strings.Split(outStr, "\nFound KubeArmor running in Kubernetes\n\n")
```
Aryan-sharma11, line 208: why are we doing this?

Manik2708: There were some absurds strings getting printed like `getting logs from pod name` and after your suggestion I thought of removing it.

Aryan-sharma11: imo we shouldn't handle it like this, it might require flushing the os.stdout before writing the probe output or may be locks so all the go routines don't write in stdout at the same time.  @Prateeknandle @DelusionalOptimist  wdyt ?

Manik2708: I also had some doubt in handling concurrency. I think using locks would be helpful but then we have to add them everywhere else in the codebase. But we can flush OS.Stdout for sure!

Manik2708: So after the meet we got to the conclusion that we should not use locks as it might block other go routines! What I have thought as a solution is using a different writer by creating and passing it in `print` function. It will fix that as no longer we will have to depend on `os.Stdout`

Aryan-sharma11 (approved):

Aryan-sharma11:
please squash the commits too.

DelusionalOptimist (commented):
@Manik2708 Please rebase with main.

Aryan-sharma11 (approved):

probe/print.go
```diff
@@ -25,87 +28,118 @@ func renderOutputInTableWithNoBorders(data [][]string) {
...
 }
 
-// printDaemonsetData function
-func printDaemonsetData(daemonsetStatus *Status) {
-	var data [][]string
+func (o *Options) getPrintableString(c *color.Color, s string) string {
+	if o.Output == "no-color" || c == nil {
+		return s
+	} else {
+		return c.SprintFunc()(s)
```
Prateeknandle, line 35: @Manik2708 can you please handle this warning here

Prateeknandle:
please clear the go-lint issue and then we'll merge this.
Thanks fro the PR @Manik2708

Prateeknandle (approved):

PR #465, opened 2024-10-25
cmd(profile): improve karmor profile for unorchestrated mode

tesla59:
part of https://github.com/kubearmor/KubeArmor/issues/1815

### DIFF
--- a/cmd/profile.go
+++ b/cmd/profile.go
@@ -4,11 +4,12 @@
 package cmd
 
 import (
+	"github.com/kubearmor/kubearmor-client/profile"
 	profileclient "github.com/kubearmor/kubearmor-client/profile/Client"
 	"github.com/spf13/cobra"
 )
 
-var profileOptions profileclient.Options
+var profileOptions profile.Options
 
 // profileCmd represents the profile command
 var profilecmd = &cobra.Command{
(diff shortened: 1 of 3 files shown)


tesla59:
@DelusionalOptimist @daemon1024 
Please review. This is first PR for the task for improving `karmor profile` for unchestrated mode. Includes
- New flags LogType and LogFilter for `karmor profile`
- Fixed the bug where `karmor profile` failed if grpc is not set in unorchestrated mode
- Dynamically set the view of profile TUI based on the passed LogType

Next task includes:
- Dynamically change the view from the TUI using a key(TBD)

rootxrishabh (commented):
Ignore this message, was meant for another PR xD

tesla59:
@Prateeknandle yes it seem to be working together. Although I am not entirely sure how the logFilter=policy would work but judging from the output of setting it to system, policy and all, it looks like it is working

Also profile starts the observer with both flags passed so it should work as expected
https://github.com/kubearmor/kubearmor-client/pull/465/files#diff-213323d4ae136c7b733df56c9658661e0b6b6aa64f60adc7ca95b9e3a04ec367R84

Prateeknandle:
@tesla59 consider these points:
- simply executing `karmor profile` is not showing host logs, I only see container logs
- `--logFilter=policy` is not working, it should show policy based logs i.e. alerts, you can check this functionality in `karmor logs --logFilter=policy `

Issue #270, opened 2023-02-16
support karmor tool on windows


nyrahul:
Currently, karmor tool can be installed only from linux. It should be possible to install kubearmor using karmor client tool on windows.

Tasks:
- [ ] Code change to compile on windows
- [ ] karmor client release for windows. Release process change.
- [ ] Installation command change.

daemon1024:
Ref #50

akashsawan1:
Hey @nyrahul ,I would like to take this one on

nyrahul:
> Hey @nyrahul ,I would like to take this one on

Great. Thank you for your interest.

akashsawan1:
Could you help me with some resource to get started.

nyrahul:
> Could you help me with some resource to get started.

KubeArmor uses goreleaser to get the releases prepared for Mac and Linux. You will have to enable windows in there and check if the compilation etc works fine. I guess you will run into issues that will need fixes.

khareyash05:
Hey @nyrahul . Is this issue open to contribute?

nyrahul:
Yes, this is still up as a good-first-issue and we welcome contributions
towards this. Thanks

EraKin575:
@nyrahul is this issue still relevant?

nyrahul:
> @nyrahul is this issue still relevant?

Yes imo it is still relevant. Cc: @daemon1024

EraKin575:
please assign me this issue

daemon1024:
@EraKin575 thanks for taking it up. I see you have asked to be assigned on a couple of issues, let's focus on other issues once we wrap this up. Appreciate the enthusiasm!

Manik2708:
Would like to work on it, please assign @daemon1024 @nyrahul

Manik2708:
@nyrahul So I tried compiling the application for windows. I got errors from following packaged:
1) Probe: This package uses `sys/unix`, so I seperated the logic of printing probe data for windows and unix where unix will print data even if the enviornment is not k8s but in windows if k8s enviornment is not found then it will throw error.
2) VM: This package uses cilium which can't be compiled on windows, so every command related to this package will not be available on windows.
3) Profile: This package uses an accunox package which is using cilium package, so again this command can't be made available for windows.
@nyrahul @daemon1024 Please see if these commands can be removed for windows build. Or are there some alternatives?

Manik2708:
[...]

Issue #50, opened 2022-03-15
Error compiling kArmor for Windows


daemon1024:
Recent cilium dep introduced in #45 is causing issues cross compiling kArmor for the Windows platform.

Error log can be found at https://github.com/kubearmor/kubearmor-client/runs/5549174379?check_suite_focus=true

To reproduce the issue run:
```sh
 env GOOS=windows CGO_ENABLED=0 go build -o karmor
```

PR #471, opened 2024-12-09
Karmor support for windows and removal of cilium

Manik2708:
Fixes: #270, #50 
Karmor is now succesfully getting compiled on windows with removal of cilium completely from repository!


Manik2708:
Please review @daemon1024 @nyrahul

profile/pathAggregator.go
```diff
@@ -0,0 +1,341 @@
+// This code is directly taken from github.com/accuknox/auto-policy-discovery/src/common
```
Aryan-sharma11, line 1: can we import this package instead of rewriting it? Is it modified according to karmor probe?

Manik2708: The reason is: The accunox packages are dependent on cilium but the part of code which we require is independent of cilium. Using the package will introduce cilium dependency which will make profiling unavailable for windows as it can't be compiled on windows!

profile/libs.go
```diff
@@ -0,0 +1,21 @@
+// This code is directly taken from github.com/accuknox/auto-policy-discovery/src/common
+
+package profile
+
+import "reflect"
+
+func ContainsElement(slice interface{}, element interface{}) bool {
```
Aryan-sharma11, line 7: let's use the official go package. https://pkg.go.dev/slices#Contains

Aryan-sharma11 (changes requested):

Manik2708:
Thanks for the commits @Aryan-sharma11, can you please tell me what is causing these security check failures?

Aryan-sharma11:
Tested on windows working as expected 
image
image
image

Aryan-sharma11:
LGTM .

Aryan-sharma11 (approved):

daemon1024 (approved):

PR #485, opened 2025-02-24
`annotateExisting` for karmor install

Aryan-sharma11:
https://github.com/kubearmor/KubeArmor/pull/1952

### DIFF
--- a/cmd/install.go
+++ b/cmd/install.go
@@ -68,6 +68,7 @@ func init() {
 	installCmd.Flags().BoolVar(&installOptions.AlertThrottling, "alertThrottling", true, "Enable/Disable Alert Throttling, by default it's enabled")
 	installCmd.Flags().Int32Var(&installOptions.MaxAlertPerSec, "maxAlertPerSec", 10, "Maximum number of alerts required to trigger alert throttling")
 	installCmd.Flags().Int32Var(&installOptions.ThrottleSec, "throttleSec", 30, "Time window(in sec) for which there will be no alerts genrated after alert throttling is triggered")
+	installCmd.Flags().BoolVar(&installOptions.AnnotateExisting, "annotateExisting", false, "when true kubearmor-controller restarts and annotates existing resources, with required annotations")
 	markDeprecated(installCmd, "env", "Only relevant when using legacy")
 	markDeprecated(installCmd, "legacy", "KubeArmor now utilizes operator-based installation. This command may not set up KubeArmor in the intended way.")
 }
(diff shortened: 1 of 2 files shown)


cmd/install.go
```diff
@@ -68,6 +68,6 @@ func init() {
 	installCmd.Flags().BoolVar(&installOptions.AlertThrottling, "alertThrottling", true, "Enable/Disable Alert Throttling, by default it's enabled")
 	installCmd.Flags().Int32Var(&installOptions.MaxAlertPerSec, "maxAlertPerSec", 10, "Maximum number of alerts required to trigger alert throttling")
 	installCmd.Flags().Int32Var(&installOptions.ThrottleSec, "throttleSec", 30, "Time window(in sec) for which there will be no alerts genrated after alert throttling is triggered")
+	installCmd.Flags().BoolVar(&installOptions.AnnotateExisting, "annotateExisting", false, "when true kubearmor-controller restarts and annotates existing resources, with required annotations")
 	markDeprecated(installCmd, "env", "Only relevant when using legacy")
-	markDeprecated(installCmd, "legacy", "KubeArmor now utilizes operator-based installation. This command may not set up KubeArmor in the intended way.")
```
daemon1024, line 72: why are we removing legacy?

install/install.go
```diff
@@ -905,7 +922,8 @@ func getOperatorConfig(o Options) map[string]interface{} {
 				"repository": operatorImage,
 				"tag":        operatorImageTag,
 			},
-			"imagePullPolicy": operatorImagePullPolicy,
+			"imagePullPolicy":  operatorImagePullPolicy,
+			"annotateExisting": o.AnnotateExisting,
```
daemon1024, line 926: can we verify if we overwrite operator config if an existing CR already exists?

install/install.go
```diff
@@ -284,6 +285,22 @@ func checkPods(c *k8s.Client, o Options, i bool) {
 			break
 		}
 	}
+	// add annotation for apparmor
+	if !o.AnnotateExisting {
```
daemon1024, line 289: We should verify this post KubeArmor starts running I guess? This way we are sure that KubeArmor is running as well and only then we show the message?

cmd/install.go
Aryan-sharma11: > why are we removing legacy?

my bad, removed it by mistake

rksharma95:
working as expected :+1: 

image

daemon1024 (approved):

daemon1024 (approved):

rksharma95 (approved):

rootxrishabh (approved):
Working as expected 🚀

PR #492, opened 2025-05-14
add log source column in profile

Aryan-sharma11:
Add log source column in karmor profile to differentiate between container logs and host logs.
image

### DIFF
--- a/profile/Client/profileClient.go
+++ b/profile/Client/profileClient.go
@@ -419,6 +419,8 @@ func generateRowsFromData(data []pb.Log, Operation string) []table.Row {
 				var logType string
 				if entry.Type == "HostLog" {
 					logType = "Host"
+					entry.NamespaceName = "--"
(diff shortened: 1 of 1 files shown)


profile/Client/profileClient.go
```diff
@@ -411,9 +416,16 @@ func generateRowsFromData(data []pb.Log, Operation string) []table.Row {
 				(entry.ContainerName == o1.Container) ||
 				(len(o1.Namespace) == 0 && len(o1.Pod) == 0 && len(o1.Container) == 0) {
 				var p Profile
+				var logType string
```
PrimalPimmy, line 419: Maybe we can add in this condition for namespace and container_name? Since they don't exist in Host logs, we can just put them as "--" instead of empty space

PrimalPimmy (commented):
A small thing, rest LGTM

PrimalPimmy (approved):

PR #494, opened 2025-05-21
minor fix

Aryan-sharma11:


rootxrishabh (approved):

install/install.go
```diff
@@ -124,7 +124,13 @@ func getOperatorCR(o Options) (*Operatorv1.KubeArmorConfig, string) {
 	var defaultCapabilitiesPosture Operatorv1.PostureType
 	var defaultNetworkPosture Operatorv1.PostureType
 
+	if len(o.KubearmorImage) == 0 {
+		o.KubearmorImage = "kubearmor/kubearmor:stable"
+	}
 	o.KubearmorImage = updateImageTag(o.KubearmorImage, o.KubeArmorTag)
+	if o.InitImage == "" {
+		o.InitImage = "kubearmor/kubearmor-init:stable"
+	}
```
rksharma95, line 133: let's handle the default image values here https://github.com/kubearmor/kubearmor-client/blob/302079178b6c5fd3578bc1f49d02305818d9f496/cmd/install.go#L89-L90

rksharma95 (changes requested):

install/install.go
Aryan-sharma11: This was done due to non-K8s installation. For systemd installation, we cannot use the stable or latest tags; it only works with specific KubeArmor release version tags.

rksharma95: can't we handle that some other way? if not please add a comment here for future reference.

rksharma95 (approved):

rootxrishabh (approved):

rksharma95 (approved):

PR #498, opened 2025-06-11
Reduce CPU usage in karmor profile

harisudarsan1:
#### Summary of Changes

This PR introduces significant performance improvements and refactors the Karmor profile view to reduce unnecessary CPU consumption. The key changes include:

1. **Removal of Periodic Event Loop and Mutexes**
   Eliminates the background event loop that was running every 2 seconds and updating the UI regardless of new events, leading to unnecessary CPU usage. Also removes the use of mutexes by streamlining the data flow.

2. **Event-Driven Updates**
   UI updates are now triggered only when a new event arrives via the `EventChan` channel, ensuring updates occur only when necessary.

3. **Operation-Specific Row Generation**
   Rows and columns are now generated only if the incoming event matches the expected operation type, avoiding redundant processing.

4. **Direct and Efficient Data Handling**
   Instead of iterating over existing logs and filtering them, the system now directly updates existing data or appends new rows if the event is unique—resulting in faster and more efficient log handling.

5. **On-Demand Export to JSON**
   Table data can now be exported as JSON by pressing `e`, instead of exporting every 2 seconds as part of the now-removed periodic update loop.

6. **Codebase Refactoring**
   Cleans up and reorganizes the code for better readability, maintainability, and performance.

### DIFF
--- a/profile/Client/helpers.go
+++ b/profile/Client/helpers.go
@@ -89,7 +89,7 @@ func ExportRowsToJSON(columns []table.Column, rows []table.Row, operation string
 
 	// Create ProfileSummary directory if it doesn't exist
 	outputDir := filepath.Join(cwd, "ProfileSummary")
-	if err := os.MkdirAll(outputDir, 0755); err != nil {
+	if err := os.MkdirAll(outputDir, 0700); err != nil {
 		return "", fmt.Errorf("failed to create output directory: %w", err)
 	}
 
@@ -98,7 +98,7 @@ func ExportRowsToJSON(columns []table.Column, rows []table.Row, operation string
 	filePath := filepath.Join(outputDir, fileName)
 
 	// Write JSON to file
-	if err := os.WriteFile(filePath, jsonBytes, 0644); err != nil {
+	if err := os.WriteFile(filePath, jsonBytes, 0600); err != nil {
 		return "", fmt.Errorf("failed to write file: %w", err)
 	}
 


profile/Client/profileClient.go
```diff
@@ -179,6 +192,9 @@ func (m Model) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 		switch {
 		case key.Matches(msg, m.keys.Quit):
 			m.quitting = true
+			if ProfileOpts.Save {
+				// export to json
```
rksharma95, line 196: is anything pending here?

profile/Client/profileClient.go
```diff
@@ -317,204 +444,12 @@ type Profile struct {
...
 // Start entire TUI
 func Start(o Options) {
-	o1 = Options{
-		Namespace: o.Namespace,
-		Pod:       o.Pod,
-		GRPC:      o.GRPC,
-		Container: o.Container,
-		Save:      o.Save,
-	}
+	ProfileOpts = o
```
rksharma95, line 449: let's directly use input o

profile/Client/helpers.go
```diff
@@ -0,0 +1,110 @@
...
+	if err := os.MkdirAll(outputDir, 0755); err != nil {
+		return "", fmt.Errorf("failed to create output directory: %w", err)
+	}
+
+	// Define the output file path
+	fileName := fmt.Sprintf("%s.json", operation)
+	filePath := filepath.Join(outputDir, fileName)
+
+	// Write JSON to file
+	if err := os.WriteFile(filePath, jsonBytes, 0644); err != nil {
```
PrimalPimmy, line 101: This might be failing Go sec checks.

harisudarsan1:
@PrimalPimmy verified with karmor logs having same number of logs

PrimalPimmy (approved):

rksharma95 (approved):

PR #561, opened 2026-09-03
Separate Management/Log PKI for unorchestrated mode

sagarkhandagre998:
Generate two independent trust planes (management + log), each with its own CA and karmor client identity, reusing the existing GenerateCA and SignCSR helpers from install/pki.go. Server certificates stay ephemeral via the server SelfSignedCertLoader; only ca.crt/ca.key and client.crt/client.key are persisted per plane.

Wire credential selection: vm policy commands use the management plane over mTLS (default localhost:32765), log commands prefer the log plane directory with single-CA fallback. Hook PKI generation into karmor install --nonk8s (docker + systemd) via SetupVMTLS.

```

                              karmor vm init
                                    │
                  ┌─────────────────┴─────────────────┐
                  │                                   │
                  ▼                                   ▼
          MANAGEMENT CA                         LOG CA
          ca.cert + ca.key                      ca.cert + ca.key
                  │                                   │
          ┌───────┴───────┐                   ┌───────┴───────┐
          │               │                   │               │
          ▼               ▼                   ▼               ▼
     Server Cert      Client Cert        Server Cert      Client Cert
     (ephemeral)      (persistent)       (ephemeral)      (persistent)
          │               │                   │               │
          │               │                   │               │
          ▼               │                   ▼               │
   Management Server      │              Log Server           │
          │               │                   │               │
          │               │                   │               │
          └─────── mTLS ──┘                   └────── mTLS ───┘
                  │                                   │
                  │                                   │
             karmor management                   karmor logs
```

### DIFF
--- a/cmd/log.go
+++ b/cmd/log.go
@@ -67,7 +67,7 @@ func init() {
 
 	logCmd.Flags().StringVar(&logOptions.GRPC, "gRPC", "", "gRPC server information")
 	logCmd.Flags().BoolVar(&logOptions.Secure, "secure", false, "connect to kubearmor on a secure connection")
-	logCmd.Flags().StringVar(&logOptions.TlsCertPath, "tlsCertPath", "/var/lib/kubearmor/tls", "path to the ca.crt, client.crt, and client.key if certs are provided locally")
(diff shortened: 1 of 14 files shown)


sagarkhandagre998:
@Aryan-sharma11 Please take a look.

sagarkhandagre998:
Hi @Aryan-sharma11 ,

This PR currently covers cert generation + karmor side only. The agent side (compose mounts, TLS flags) would flip existing setups from plaintext to mTLS. Should that go into this same PR, or a follow-up?

I think we should include in this itself , WDYT?

